PR #1, opened 2017-01-17
Sync with upstream

torvalds:
This merges the current libdivecomputer upstream code.

Note the new DC_SAMPLE_GASMIX thing - I'm sending another pull request for subsurface for that


GoldenVadim:
first

PR #6, opened 2017-05-24
Sync with Jef's upstream

janmulder:
Merges changes from Jef's libdc upstream into the Subsurface-branch

Add support for the new models (ix3m)
Add support for the new iX3M APOS4 firmware
Add the firmware version to the devinfo event (ix3m)
Report the ppO2 in external O2 sensor mode only (Shearwater)
Replace the constant offset with a scaling factor (Shearwater)
Apply the calibration correction only for the Predator
shearwater: Report individual sensor values
Predator: don't report PPO2 unless in CC mode
Fix the Sherwood Wisdom 3 memory layout
Implement the read and write functions for the OSTC3
Add support for the Mares Quad

In addition some merge conflicts are resolved.

Signed-off-by: Jan Mulder <jlmulder@xs4all.nl>


torvalds:
On Wed, May 24, 2017 at 10:13 AM, janmulder <notifications@github.com> wrote:
>
> In addition some merge conflicts are resolved.

So since I want to know what's going on, I ended up doing my own
merging and re-resolving things, and while I did that I became
convinced that there is a bug in the PPO2 reporting for the predator.

The code does this:

#ifdef SENSOR_AVERAGE
                        sample.ppo2 = data[offset + 6] / 100.0;
                        if (callback) callback (DC_SAMPLE_PPO2,
sample, userdata);
#else
                        if ((status & PPO2_EXTERNAL) == 0) {
                                sample.ppo2 = data[offset + 12] *
parser->calibration[0];
                                if (callback && (data[86] & 0x01))
callback (DC_SAMPLE_PPO2, sample, userdata);

                                sample.ppo2 = data[offset + 14] *
parser->calibration[1];
                                if (callback && (data[86] & 0x02))
callback (DC_SAMPLE_PPO2, sample, userdata);

                                sample.ppo2 = data[offset + 15] *
parser->calibration[2];
                                if (callback && (data[86] & 0x04))
callback (DC_SAMPLE_PPO2, sample, userdata);
                        }
#endif

which means that if you don't have SENSOR_AVERAGE defined, and if it's
an external sensor, now there's no sensor reporting AT ALL.

Which seems bogus.

So either that average PPO2 should be sent if there's an external
sensor, or that SENSOR_AVERAGE case should also be inside that "check
for external sensor". The code as it stands now makes no sense.

Anton, I think this is all yours. Comments?

                 Linus

torvalds:
On Thu, May 25, 2017 at 10:21 AM, Linus Torvalds
<torvalds@linux-foundation.org> wrote:
>
> So since I want to know what's going on, I ended up doing my own
> merging and re-resolving things, and while I did that I became
> convinced that there is a bug in the PPO2 reporting for the predator.

Side note: I pushed out the merge anyway, because the bug is not in
the merge, it's in the original code. But let's aim to fix it.

Jan, can you check my merge against yours? I noticed an older
(harmless) merge error wrt the Shearwater sample size parser state
initialization, so our merges may not be identical, but as usual I
don't actually have any of the devices that the code changes touch, so
it would be good to check.

(Actually, I guess I have Dirk's Shearwater for my failed BT testing,
I could try to see if I can get at least the old one working)

               Linus

torvalds:
Closing pull request - it now conflicts because of slightly different resolution.

I resolved our shearwater "serial" argument addition by putting it at the end of the upstream arguments, while Jan resolved it the other way around.

janmulder:
Perfectly happy with your merge Linus. 

Thanks for processing this, as we have (at least) two users with iX3M computers that reported the issue with the APOS4 firmware.

torvalds:
On 25-05-17 19:21, Linus Torvalds wrote:
> On Wed, May 24, 2017 at 10:13 AM, janmulder <notifications@github.com> wrote:
>>
>> In addition some merge conflicts are resolved.
>
> So since I want to know what's going on, I ended up doing my own
> merging and re-resolving things, and while I did that I became
> convinced that there is a bug in the PPO2 reporting for the predator.
>
> The code does this:
>
> #ifdef SENSOR_AVERAGE
>                         sample.ppo2 = data[offset + 6] / 100.0;
>                         if (callback) callback (DC_SAMPLE_PPO2,
> sample, userdata);
> #else
>                         if ((status & PPO2_EXTERNAL) == 0) {
>                                 sample.ppo2 = data[offset + 12] *
> parser->calibration[0];
>                                 if (callback && (data[86] & 0x01))
> callback (DC_SAMPLE_PPO2, sample, userdata);
>
>                                 sample.ppo2 = data[offset + 14] *
> parser->calibration[1];
>                                 if (callback && (data[86] & 0x02))
> callback (DC_SAMPLE_PPO2, sample, userdata);
>
>                                 sample.ppo2 = data[offset + 15] *
> parser->calibration[2];
>                                 if (callback && (data[86] & 0x04))
> callback (DC_SAMPLE_PPO2, sample, userdata);
>                         }
> #endif
>
> which means that if you don't have SENSOR_AVERAGE defined, and if it's
> an external sensor, now there's no sensor reporting AT ALL.
>
> Which seems bogus.
>
> So either that average PPO2 should be sent if there's an external
> sensor, or that SENSOR_AVERAGE case should also be inside that "check
> for external sensor". The code as it stands now makes no sense.
>
> Anton, I think this is all yours. Comments?

The explanation is in the commit message:

     Note that the interpretation of the PPO2 status bit appears to be
     reversed (0=external and 1=internal).

I think this is a mistake in the Shearwater documentation. I verified against 
data from a few users: for fixed setpoint dives the bit is set and for dives 
with external O2 sensors it's not set.

Does it make sense now?

Jef

torvalds:
On Thu, May 25, 2017 at 11:35 AM, Jef Driesen <jef@libdivecomputer.org> wrote:
>
>
> The explanation is in the commit message:
>
>     Note that the interpretation of the PPO2 status bit appears to be
>     reversed (0=external and 1=internal).

You mis-understand. It's not the particular status bit polarity I
worry about. It's simply that whole "average PPO2 data" piece.

IOW, this code:

  #ifdef SENSOR_AVERAGE
        sample.ppo2 = data[offset + 6] / 100.0;
        if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
  #else

fundamentally makes no sense as it is now.

There are two possibilities I see for that sample.ppo2 value:

 (a) it too only exists for the internal sensor case

 (b) it exists even when we have an external sensor, and then contains
the single PPO2 value that the external sensor gives us.

but the code right now is wrong in *either* case.

So if it's case (a), then that whole #ifdef SENSOR_AVERAGE logic
should be _inside_ the PPO2_EXTERNAL check.

And if it is case (b), then we should have an "else" case for that
PPO2_EXTERNAL case, and give the user that single PPO2 value.

See what I'm saying? As it is, the code makes no sense and absolutely
cannot be right.

                 Linus

torvalds:
On 25-05-17 20:53, Linus Torvalds wrote:
> On Thu, May 25, 2017 at 11:35 AM, Jef Driesen <jef@libdivecomputer.org> wrote:
>>
>>
>> The explanation is in the commit message:
>>
>>     Note that the interpretation of the PPO2 status bit appears to be
>>     reversed (0=external and 1=internal).
>
> You mis-understand. It's not the particular status bit polarity I
> worry about. It's simply that whole "average PPO2 data" piece.
>
> IOW, this code:
>
>   #ifdef SENSOR_AVERAGE
>         sample.ppo2 = data[offset + 6] / 100.0;
>         if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
>   #else
>
> fundamentally makes no sense as it is now.
>
> There are two possibilities I see for that sample.ppo2 value:
>
>  (a) it too only exists for the internal sensor case
>
>  (b) it exists even when we have an external sensor, and then contains
> the single PPO2 value that the external sensor gives us.
>
> but the code right now is wrong in *either* case.
>
> So if it's case (a), then that whole #ifdef SENSOR_AVERAGE logic
> should be _inside_ the PPO2_EXTERNAL check.
>
> And if it is case (b), then we should have an "else" case for that
> PPO2_EXTERNAL case, and give the user that single PPO2 value.
>
> See what I'm saying? As it is, the code makes no sense and absolutely
> cannot be right.

Yes, now I see what you mean. At the moment, if SENSOR_AVERAGE is defined, we 
always report a ppO2 value, and when it's not defined we only report ppO2 values 
in external O2 sensor mode (but not in fixed setpoint mode). To be consistent it 
should be like this:

if ((status & PPO2_EXTERNAL) == 0) {
#ifdef SENSOR_AVERAGE
     sample.ppo2 = data[offset + 6] / 100.0;
     if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
#else
     sample.ppo2 = data[offset + 12] * parser->calibration[0];
     if (callback && (data[86] & 0x01)) callback (DC_SAMPLE_PPO2, sample, userdata);

     sample.ppo2 = data[offset + 14] * parser->calibration[1];
     if (callback && (data[86] & 0x02)) callback (DC_SAMPLE_PPO2, sample, userdata);

     sample.ppo2 = data[offset + 15] * parser->calibration[2];
     if (callback && (data[86] & 0x04)) callback (DC_SAMPLE_PPO2, sample, userdata);
#endif
}

I'll fix that, even if it doesn't really matter. (The purpose of the #ifdef is 
to comment out this code. It's not supposed to ever be enabled. The only reason 
why we didn't remove it completely is because in the future we would like to 
report both the values from the O2 sensor(s) and the average ppO2. But at the 
moment we can't distinguish the different types, and thus the average ppO2 is 
disabled. So that code is just there as a reminder.)

Whether we should report ppO2 in fixed setpoint mode (the "else" case in your 
explanation) is another question. I'm not really sure what's the right answer there.

Jef

torvalds:
On 27 May, 2017 - Jef Driesen wrote:

> On 25-05-17 20:53, Linus Torvalds wrote:
> >On Thu, May 25, 2017 at 11:35 AM, Jef Driesen <jef@libdivecomputer.org> wrote:
> >>
> >>
> >>The explanation is in the commit message:
> >>
> >>    Note that the interpretation of the PPO2 status bit appears to be
> >>    reversed (0=external and 1=internal).
> >
> >You mis-understand. It's not the particular status bit polarity I
> >worry about. It's simply that whole "average PPO2 data" piece.
> >
> >IOW, this code:
> >
> >  #ifdef SENSOR_AVERAGE
> >        sample.ppo2 = data[offset + 6] / 100.0;
> >        if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
> >  #else
> >
> >fundamentally makes no sense as it is now.
> >
> >There are two possibilities I see for that sample.ppo2 value:
> >
> > (a) it too only exists for the internal sensor case
> >
> > (b) it exists even when we have an external sensor, and then contains
> >the single PPO2 value that the external sensor gives us.
> >
> >but the code right now is wrong in *either* case.
> >
> >So if it's case (a), then that whole #ifdef SENSOR_AVERAGE logic
> >should be _inside_ the PPO2_EXTERNAL check.
> >
> >And if it is case (b), then we should have an "else" case for that
> >PPO2_EXTERNAL case, and give the user that single PPO2 value.
> >
> >See what I'm saying? As it is, the code makes no sense and absolutely
> >cannot be right.
> 
> Yes, now I see what you mean. At the moment, if SENSOR_AVERAGE is
> defined, we always report a ppO2 value, and when it's not defined we
> only report ppO2 values in external O2 sensor mode (but not in fixed
> setpoint mode). To be consistent it should be like this:
> 
> if ((status & PPO2_EXTERNAL) == 0) {
> #ifdef SENSOR_AVERAGE
>     sample.ppo2 = data[offset + 6] / 100.0;
>     if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
> #else
>     sample.ppo2 = data[offset + 12] * parser->calibration[0];
>     if (callback && (data[86] & 0x01)) callback (DC_SAMPLE_PPO2, sample, userdata);
> 
>     sample.ppo2 = data[offset + 14] * parser->calibration[1];
>     if (callback && (data[86] & 0x02)) callback (DC_SAMPLE_PPO2, sample, userdata);
> 
>     sample.ppo2 = data[offset + 15] * parser->calibration[2];
>     if (callback && (data[86] & 0x04)) callback (DC_SAMPLE_PPO2, sample, userdata);
> #endif
> }
> 
> I'll fix that, even if it doesn't really matter. (The purpose of the
> #ifdef is to comment out this code. It's not supposed to ever be
> enabled. The only reason why we didn't remove it completely is
> because in the future we would like to report both the values from
> the O2 sensor(s) and the average ppO2. But at the moment we can't
> distinguish the different types, and thus the average ppO2 is
> disabled. So that code is just there as a reminder.)
> 
> Whether we should report ppO2 in fixed setpoint mode (the "else"
> case in your explanation) is another question. I'm not really sure
> what's the right answer there.

When I added the #ifdef SENSOR_AVERAGE thingie was just to point out the
two different code paths.

I didn't change the design, IE reporting DC_SAMPLE_PPO2 when diving
without external sensors because I thought that was out of scope for
that patch.



That said. I agree, reporting DC_SAMPLE_PPO2 when running in fixed
setpoint mode without external sensors is just confusing at best, and we
should stop doing that. We should only report a DC_SAMPLE_SETPOINT
because thats what the computer actually uses.

Good catch there Linus.

Some users might be confused when the "sensor" disappears when they are
running their computer in fixed setpoint mode, but it should be easy to
explain.


//Anton


-- 
Anton Lundin	+46702-161604

torvalds:
On 2017-05-29 09:16, Anton Lundin wrote:
> On 27 May, 2017 - Jef Driesen wrote:
> 
>> On 25-05-17 20:53, Linus Torvalds wrote:
>> >On Thu, May 25, 2017 at 11:35 AM, Jef Driesen <jef@libdivecomputer.org> wrote:
>> >>
>> >>
>> >>The explanation is in the commit message:
>> >>
>> >>    Note that the interpretation of the PPO2 status bit appears to be
>> >>    reversed (0=external and 1=internal).
>> >
>> >You mis-understand. It's not the particular status bit polarity I
>> >worry about. It's simply that whole "average PPO2 data" piece.
>> >
>> >IOW, this code:
>> >
>> >  #ifdef SENSOR_AVERAGE
>> >        sample.ppo2 = data[offset + 6] / 100.0;
>> >        if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
>> >  #else
>> >
>> >fundamentally makes no sense as it is now.
>> >
>> >There are two possibilities I see for that sample.ppo2 value:
>> >
>> > (a) it too only exists for the internal sensor case
>> >
>> > (b) it exists even when we have an external sensor, and then contains
>> >the single PPO2 value that the external sensor gives us.
>> >
>> >but the code right now is wrong in *either* case.
>> >
>> >So if it's case (a), then that whole #ifdef SENSOR_AVERAGE logic
>> >should be _inside_ the PPO2_EXTERNAL check.
>> >
>> >And if it is case (b), then we should have an "else" case for that
>> >PPO2_EXTERNAL case, and give the user that single PPO2 value.
>> >
>> >See what I'm saying? As it is, the code makes no sense and absolutely
>> >cannot be right.
>> 
>> Yes, now I see what you mean. At the moment, if SENSOR_AVERAGE is
>> defined, we always report a ppO2 value, and when it's not defined we
>> only report ppO2 values in external O2 sensor mode (but not in fixed
>> setpoint mode). To be consistent it should be like this:
>> 
>> if ((status & PPO2_EXTERNAL) == 0) {
>> #ifdef SENSOR_AVERAGE
>>     sample.ppo2 = data[offset + 6] / 100.0;
>>     if (callback) callback (DC_SAMPLE_PPO2, sample, userdata);
>> #else
>>     sample.ppo2 = data[offset + 12] * parser->calibration[0];
>>     if (callback && (data[86] & 0x01)) callback (DC_SAMPLE_PPO2, 
>> sample, userdata);
>> 
>>     sample.ppo2 = data[offset + 14] * parser->calibration[1];
>>     if (callback && (data[86] & 0x02)) callback (DC_SAMPLE_PPO2, 
>> sample, userdata);
>> 
>>     sample.ppo2 = data[offset + 15] * parser->calibration[2];
>>     if (callback && (data[86] & 0x04)) callback (DC_SAMPLE_PPO2, 
>> sample, userdata);
>> #endif
>> }
>> 
>> I'll fix that, even if it doesn't really matter. (The purpose of the
>> #ifdef is to comment out this code. It's not supposed to ever be
>> enabled. The only reason why we didn't remove it completely is
>> because in the future we would like to report both the values from
>> the O2 sensor(s) and the average ppO2. But at the moment we can't
>> distinguish the different types, and thus the average ppO2 is
>> disabled. So that code is just there as a reminder.)
>> 
>> Whether we should report ppO2 in fixed setpoint mode (the "else"
>> case in your explanation) is another question. I'm not really sure
>> what's the right answer there.
> 
> When I added the #ifdef SENSOR_AVERAGE thingie was just to point out 
> the
> two different code paths.
> 
> I didn't change the design, IE reporting DC_SAMPLE_PPO2 when diving
> without external sensors because I thought that was out of scope for
> that patch.
> 
> 
> 
> That said. I agree, reporting DC_SAMPLE_PPO2 when running in fixed
> setpoint mode without external sensors is just confusing at best, and 
> we
> should stop doing that. We should only report a DC_SAMPLE_SETPOINT
> because thats what the computer actually uses.
> 
> Good catch there Linus.
> 
> Some users might be confused when the "sensor" disappears when they are
> running their computer in fixed setpoint mode, but it should be easy to
> explain.

I just pushed the patch to libdivecomputer master.

Jef

PR #7, opened 2017-06-22
Create ne w"custom_io" packet sending model 

torvalds:
This creates a new "custom_io" model that can be used to send and receive "packets" over some transport, currently only USBHID.

It also converts the two USBHID backends (Suunto EON Steel and Scubapro G2) to use that custom IO model.

The plan is to be able to then replace the USBHID transport with a BLE GATT-based transport fairly transparently to the libdivecomputer backends (I say "fairly", because the packetization is going to be exposed, and each dive computer backend will inevitably see some transport specifics because they have their own logic of how to encode the stream over the packet interface).

### DIFF
--- a/include/libdivecomputer/Makefile.am
+++ b/include/libdivecomputer/Makefile.am
@@ -3,7 +3,7 @@ libdivecomputer_HEADERS = 	\
 	version.h \
 	common.h \
 	context.h \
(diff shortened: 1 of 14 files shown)


dahuoyzs:
good

ghost:
I approve.

PR #10, opened 2017-11-09
OSTC3: also interface the initial CNS from header

janmulder:
Added the capability to read the initial CNS from the dive header, so that we can retrieve this, and initilize correctly when downloading dives from an OSTC.

Signed-off-by: Jan Mulder <jlmulder@xs4all.nl>


torvalds:
So I'm not convinced. A few issues:

 - why don't you just report the initial CNS in the first sample? That's what *everything* else does, and it means that there is no need for a new DC_xyz interface

 - the patch has some extra whitespace that makes no sense

janmulder:
Interestingly, the OSTC does not seem to report a CNS value every sample. I tried to understand the HW interface spec on Extendend Information, to see if this is parser limitation on our side, or just a property of the OSTC protocol. Not exactly sure about this, but tempted to say that this is just the protocol.

There is a very simple 1 line "fix" in Subsurface only. Just resetting the CNS to 0 at start a new dive (which is obviously not correct in repetitive diving), but a first report from the OSTC brings us back in a sane state. Currently, with a sample rate of 2 sec, a CNS is reported within 26 seconds. So for dives with a non zero initial CNS, the first 24 seconds is wrong ... 

This all said ... I agree that this commit is overly complex for useless data as CNS.

sfuchs79:
@janmulder 
Ah, ok, fine for me. Then let's do the cns = 0 thing at the "usual" place you already pointed to at the beginning of our conversation?
Means also correcting reference files for tests but that's not too difficult.

sfuchs79:
@janmulder @torvalds 
Meanwhile I looked at this again and I have to say that I have some doubts about Linus doubts ;-)
What is really wrong about taking the initial CNS from the dive header and putting it into our first sample(s) until we receive a sample CNS from the OSTC for the first time?
I agree that it may be a waste to spent so much time on this and I would also agree to a simpler change where we reset the cns and start with a cns of 0 in the first sample(s) but I would suggest that we get rid of this erroneous "final cns of last dive in first samples of next dive" thing.

torvalds:
On Sat, Nov 11, 2017 at 9:05 AM, Stefan Fuchs <notifications@github.com>
wrote:

> @janmulder <https://github.com/janmulder> @torvalds
> <https://github.com/torvalds>
> Meanwhile I looked at this again and I have to say that I have some doubts
> about Linus doubts ;-)
> What is really wrong about taking the initial CNS from the dive header and
> putting it into our first sample(s) until we receive a sample CNS from the
> OSTC for the first time?
>
That's what I think _should_ be done.

That's not what Jan's patch does.

Jan's patch adds a whole new FC_FIELD_xyz, which then subsurface would have
to query explicitly. So it actually expands with a totally new
libdivecomputer interface.

What I suggest is that no libdivecomputer interface changes be done, and
instead just the OSTC report the initial CNS value in the first sample
(unless it already has one).

So the change should be entirely internal to the OSTC back-end, and not
require any new interfaces, and not require any new subsurface code to
actually fetch the initial CNS.

              Linus

torvalds:
@sfuchs79 @janmulder 

Argh. Replying here too, since I don't think the github notifications work both ways over email.

Stefan Fuchs says: "What is really wrong about taking the initial CNS from the dive header and putting it into our first sample(s) until we receive a sample CNS from the OSTC for the first time?"

And indeed, that's what I think _should_ be done.

But that's not what Jan's patch does.

Jan's patch adds a whole new FC_FIELD_xyz, which then subsurface would have to query explicitly. So it actually expands with a totally new libdivecomputer interface.

What I suggest is that no libdivecomputer interface changes be done, and instead just the OSTC report the initial CNS value in the first sample (unless it already has one).

So I think the change should be entirely internal to the OSTC back-end, and not require any new interfaces, and not require any new subsurface code to actually fetch the initial CNS.

sfuchs79:
@torvalds 
Ok, I understand your point.
Where do we need to do this then? In libdc in hw_ostc_parser.c?
Don't know if @janmulder will have a look again or maybe I can also try but can't say if I will be successful with a item like this.

janmulder:
Always good to give things a rest :-) Re-re reading all this, I finally understand the way-on. I will give it another try soon.

janmulder:
New fix (all parser internal, just put the initial CNS in the first sample) is with Jef for review and incorporation in master.

janmulder:
@torvalds. Can you merge the latest libdivecomputer stuff to our branch? My fix has landed in master.

torvalds:
Merged and pushed out.

PR #14, opened 2018-03-11
Merge with upstream libdivecomputer

torvalds:
This has the cleanups to make marging easier, and then merges with Jef's upstream libdivecomputer state, giving us the new OSTC 2 TR support.

HOWEVER, there are also other changes upstream, including the removal of the old halfduplex IO interface (because the Suunto back-end that used it now just does it manually itself). And that affects subsurface (removing the halfduplex thing from the custom IO interface), so I'm sending Dirk a separate patch just for subsurface when this is merged.

The dive computer descriptor merge cleanups also means that the FTDI/USB/BT/BLE comment markers no longer exist, so Dirk needs to use something else to do the mobile support filtering.

Adding @sfuchs79 to try this out too.


dirkhh:
This took me a while to wade through.
Thanks for doing all the hard work here, Linus.

sfuchs79:
@dirkhh 
May there be now some additional steps necessary to add the "OSTC 2 TR" to some lists?
Either by running a script or by doing it manually?
Can I help somehow?
CHANGELOG is another item.

dirkhh:
I assume the OSTC 2 TR is a BLE dive computer? I'll manually add it.

sfuchs79:
@dirkhh 
I'm not 100% sure. HW website says:
Bluetooth Smart Ready (2.0 und 4.0)

BLE would mean "not possible to download dives under Win", doesn't it?

torvalds:
On Wed, Mar 14, 2018 at 10:51 AM, Stefan Fuchs <notifications@github.com> wrote:
>
> Bluetooth Smart Ready (2.0 und 4.0)

That "Smart Ready" should mean both BLE and legacy BT.

Re-inforced by that "2.0 und 4.0". It almost certainly does both.

             Linus

sfuchs79:
@torvalds 
So it "should work" under Windows?

torvalds:
On Wed, Mar 14, 2018 at 11:19 AM, Stefan Fuchs <notifications@github.com> wrote:
>
> @torvalds
> So it "should work" under Windows?

With those quotes in place, yes. It *sounds* like they just added BLE,
but that the old BT should still work.

That is how some of the previous-gen Shearwater dive computers also
worked - you could use either.

But let's see what the reality is. It sounds like nobody actually has
the 2TR yet.

I do like how it's apparently entirely wireless (including charging).
I assume that means that it lost the USB serial side, and I'm not
seeing any place to connect to on the pictures.

Which is another (good) reason to not go BLE-only. It isn't only us
who have had issues with BLE, I would expect HW to want to have a
fallback, and without USB, legacy BT is what I'd expect.

             Linus

PR #15, opened 2018-06-19
Handle Shearwater Petrel 2 Voted PO2

aquarat:
If downloading from a Shearwater dive computer and the log in question
is from a CCR dive, if calibration data for the sensors appears to be
the default (uncalibrated), emit an informative message indicating
as such (same type of message as battery voltages, serial number, etc).

If the dive computer supports a "Voted" or displayed PO2 field, retain
that field and send it off to Subsurface as a Voted PO2 field.

In the case of Shearwater dive computers, particularly the Petrel/JJ
combination, the Voted PO2 is a calibrated value as displayed to the
diver during the logged dive.

Signed-off-by: Aaron Scheiner <blue@aquarat.co.za>


aquarat:
Ergh, clearly I've made a mistake somewhere. I'll retry later. Sorry.

PR #20, opened 2020-06-25
First cut at Oceans S1 support

torvalds:
This is the first cut at Oceans S1 downloader support

It has a number of weaknesses, but the basic downloading works. Because of the "oldest dives first", you need to force download all dives so that subsurface doesn't say "oh, I already have this one" and stop downloading entirely, but I'm creating this pull request mostly as a "look, if somebody wants to test this, here's the code"

Don't actually merge upstream yet.


torvalds:
Updated to actually download in the right order and also add fingerprinting code.

dirkhh:
Should I updated Subsurface to use this?

torvalds:
On Thu, Jun 25, 2020 at 11:49 AM Dirk Hohndel <notifications@github.com> wrote:
>
> Should I updated Subsurface to use this?

No, I just sent you the proper pull request.

It's basically just the libdivecomputer update, but it's a more
complete one with a upstream merge from Jef, and it also contains the
bluetooth name recognition for the S1.

                Linus

PR #25, opened 2020-10-29
Garmin Descent Mk2/Mk2i MTP support

dirkhh:
The Descent Mk2/Mk2i no longer appears as a USB storage / FAT device, but instead as an MTP device.
Instead of adding yet another transport definition, I decided to implement this as two alternative methods of accessing the files directly in `garmin.c`.

### DIFF
--- a/src/descriptor.c
+++ b/src/descriptor.c
@@ -406,8 +406,10 @@ static const dc_descriptor_t g_descriptors[] = {
 	{"Liquivision", "Kaon", DC_FAMILY_LIQUIVISION_LYNX, 3, DC_TRANSPORT_SERIAL, NULL},
 
 	// Not merged upstream yet
-	/* Garmin */
+	/* Garmin -- model numbers as defined in FIT format; USB product id is (0x4000 | model) */
+	/* for the Mk2 we are using the model of the global model - the APAC model is 3702 */
 	{"Garmin", "Descent Mk1", DC_FAMILY_GARMIN, 2859, DC_TRANSPORT_USBSTORAGE, dc_filter_garmin},
+	{"Garmin", "Descent Mk2/Mk2i", DC_FAMILY_GARMIN, 3258, DC_TRANSPORT_USBSTORAGE, dc_filter_garmin},
 	/* Deepblu */
 	{"Deepblu", "Cosmiq+", DC_FAMILY_DEEPBLU, 0, DC_TRANSPORT_BLE, dc_filter_deepblu},
 	/* Oceans S1 */
--- a/src/garmin.c
+++ b/src/garmin.c
@@ -414,6 +414,12 @@ garmin_device_foreach (dc_device_t *abstract, dc_dive_callback_t callback, void 
 		return rc;
 	pathname[pathlen] = 0;
 
+#ifdef HAVE_LIBMTP
+	// if the user passes in a path, don't try to read via MTP
+	if (pathlen)
+		device->use_mtp = 0;
+#endif
+
 	// The actual dives are under the "Garmin/Activity/" directory
 	// as FIT files, with names like "2018-08-20-10-23-30.fit".
 	// Make sure our buffer is big enough.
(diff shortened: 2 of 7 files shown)


src/descriptor.c
```diff
@@ -406,8 +406,10 @@ static const dc_descriptor_t g_descriptors[] = {
 	{"Liquivision", "Kaon", DC_FAMILY_LIQUIVISION_LYNX, 3, DC_TRANSPORT_SERIAL, NULL},
 
 	// Not merged upstream yet
-	/* Garmin */
+	/* Garmin -- model numbers as defined in FIT format; USB product id is (0x4000 | model) */
 	{"Garmin", "Descent Mk1", DC_FAMILY_GARMIN, 2859, DC_TRANSPORT_USBSTORAGE, dc_filter_garmin},
+	{"Garmin", "Descent Mk2/Mk2i", DC_FAMILY_GARMIN, 3258, DC_TRANSPORT_USBSTORAGE, dc_filter_garmin},
+	{"Garmin", "Descent Mk2/Mk2i", DC_FAMILY_GARMIN, 3702, DC_TRANSPORT_USBSTORAGE, dc_filter_garmin},
```
torvalds: Ugh. This means that people see two different entries that look identical. That seems horribly broken.

Plus subsurface will then match on the name when doing the whole "use previously selected dive computer", and just pick the first one. So having two different entries doesn't even work, I think. No?

src/garmin.c
```diff
@@ -156,11 +186,23 @@ make_space(struct file_list *files)
...
+			unsigned int *mtp_ids;
+			new_size = n * sizeof(unsigned int);
+			mtp_ids = realloc(files->mtp_ids, new_size);
+			if (!mtp_ids)
+				return DC_STATUS_NOMEMORY;
+			files->mtp_ids = mtp_ids;
+		}
+#else
+		(void)is_mtp;
+#endif
```
torvalds: Hmm.

I'd just unconditionally add "unsigned int mtp_id" to "struct fit_name" instead. The non-mtp case can just set it to zero and ignore it.

You'd have to make the "name_cmp()" function actually do something like


    static int name_cmp(const void *_a, const void *_b)
    {
        const struct fit_name *a = _a;
        const struct fit_name *b = _b;

        // Sort reverse string ordering (newest first), so use 'b,a'
        return strcmp(b->name, a->name);
    }

and maybe there is something else that depends on "struct fit_name" only being the name, but that would seem to be a cleanup regardless.

torvalds (commented):
Small comments for what I think would be improvements, but looks sane to me.

src/descriptor.c
dirkhh: Ah, right. No, that makes no sense. I'll fix that - I learned about that second model just yesterday and of course don't have access to one (APAC only, not released,yet) so I just added that blind and didn't think it all the way through, I guess.

src/garmin.c
dirkhh: ok, I'll look into that

dirkhh:
OK, I think I have addressed your comments. Ready to merge as far as I am concerned.

torvalds (commented):
Looks fine to me, but will double-check when home

torvalds:
Ahhah. This thing is somewhat problematic, in that it does that "use_mtp" thing based on the model number of the device you've picked.

Which means that you can't select the Mk2, and then a FIT file that you've downloaded from Garmin onto your local filesystem.

I think the code should probably have a  fallback to the old USBSTORAGE code.

But I don't think this should hold you up from merging it (and merging the subsurface part - they need to go hand-in-hand because updating just libdivecomputer will then cause failures when building subsurface because it now wants the new library)

dirkhh:
Ha - that's an interesting aspect that I had not taken into account.
I think the answer is that the Garmin code should assume MTP if there is no path given, but file based if the user passes in a path. I'll take a look at that.

torvalds (approved):
This looks reasonable. Feel free to merge it all up, and I'll do the air integration today.

dirkhh:
> This looks reasonable. Feel free to merge it all up, and I'll do the air integration today.

will do. I'm using the holes in my schedule to find time to work on this :-)

PR #26, opened 2020-10-31
fix incorrect comparison

dirkhh:
Operator precedence gets me every time. Equally binds stronger
than bitwise logical operation.

### DIFF
--- a/src/garmin.c
+++ b/src/garmin.c
@@ -98,7 +98,7 @@ garmin_device_open (dc_device_t **out, dc_context_t *context, dc_iostream_t *ios
 	// for a Descent Mk2/Mk2i, we have to use MTP to access its storage;
 	// for Garmin devices, the model number corresponds to the lower three nibbles of the USB product ID
 	// in order to have only one entry for the Mk2, we don't use the Mk2/APAC model number in our code
-	device->use_mtp = (model == 0x0FFF & DESCENT_MK2);
+	device->use_mtp = (model == (0x0FFF & DESCENT_MK2));
 	DEBUG(context, "Found Garmin with model 0x%x which is a %s\n", model, (device->use_mtp ? "Mk2/Mk2i" : "Mk1"));
 #endif
 
--- a/src/garmin.c
+++ b/src/garmin.c
@@ -322,7 +322,8 @@ mtp_get_file_list(dc_device_t *abstract, struct file_list *files)
 	free(rawdevices);
 	DEBUG(abstract->context, "Found %d files", files->nr);
 
-	qsort(files->array, files->nr, sizeof(struct fit_file), name_cmp);
+	if (files->array)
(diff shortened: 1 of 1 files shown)


glance-:
Shouldn't there be some CID-references in the commit messages to?

src/garmin.c
```diff
@@ -322,7 +322,8 @@ mtp_get_file_list(dc_device_t *abstract, struct file_list *files)
 	free(rawdevices);
 	DEBUG(abstract->context, "Found %d files", files->nr);
 
-	qsort(files->array, files->nr, sizeof(struct fit_file), name_cmp);
+	if (files->array)
+		qsort(files->array, files->nr, sizeof(struct fit_file), name_cmp);
```
torvalds, line 326: Does this matter?

If the number of entries is zero, there should never be any access to the array pointer. So the test for NULL seems entirely unnecessary.

src/garmin.c
```diff
@@ -98,7 +98,7 @@ garmin_device_open (dc_device_t **out, dc_context_t *context, dc_iostream_t *ios
 	// for a Descent Mk2/Mk2i, we have to use MTP to access its storage;
 	// for Garmin devices, the model number corresponds to the lower three nibbles of the USB product ID
 	// in order to have only one entry for the Mk2, we don't use the Mk2/APAC model number in our code
-	device->use_mtp = (model == 0x0FFF & DESCENT_MK2);
+	device->use_mtp = (model == (0x0FFF & DESCENT_MK2));
```
torvalds, line 101: Ack

torvalds (approved):
Go ahead and merge.

Subsurface-divelog:
There should be, I had only a few minutes before I had to run another errand and forgot.

src/garmin.c
dirkhh: It's not entirely clear from the standard if `qsort()` might dereference the pointer if the second argument is zero.
So this seems like very cheap insurance.

dirkhh: but I guess then I should check in `get_file_list()` as well?
meh

torvalds:
On Sat, Oct 31, 2020 at 3:04 PM Dirk Hohndel <notifications@github.com> wrote:
>
> It's not entirely clear from the standard if qsort() might dereference the pointer if the second argument is zero

I can see no possible way that that wouldn't be a HUGE bug in qsort.

Basically, if qsort accesses the pointer, then it would by definition
access outside the array you gave it. That's completely invalid
behavior.

So i think that extra test isn't insurance, it's just "qsort on
platform XYZ is hugely buggy", and would like to see the actual reason
for it explained.

                Linus

dirkhh:
my apologies - I literally just merged it and added a corresponding update to upstream master.
feel free to revert

PR #29, opened 2021-08-20
More serial

dirkhh:
verified the serial data for two dive computer families

I also understand better the serial number situation for the ancient Mares Puck. There only the last four digits of the serial are in the log (so the device will show S#09008022 - but in the download data we only have "8022"). I was torn whether to add this as a string... given how old these dive computers are, I decided not to do it for now.

I also have two different Oceanic dive computers and was hoping to test whether the serial that is parsed in that device code represents what the owner can verify on device - but I appear to have misplaced my Oceanic download cable... I know I had this a few months ago, but it's not where it should be - so I'm submitting the two that I was able to already verify.

### DIFF
--- a/src/suunto_vyper_parser.c
+++ b/src/suunto_vyper_parser.c
@@ -20,6 +20,8 @@
  */
 
 #include <stdlib.h>
+#include <stdio.h>	/* for snprintf */
+#include <string.h>	/* for strdup */
 
 #include <libdivecomputer/units.h>
 
@@ -41,6 +43,7 @@ struct suunto_vyper_parser_t {
 	unsigned int maxdepth;
 	unsigned int marker;
 	unsigned int ngasmixes;
+	unsigned int serial;
 	unsigned int oxygen[NGASMIXES];
 };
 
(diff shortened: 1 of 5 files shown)


dirkhh:
I have found the Oceanic cable and was able to verify with two different models that the Atom2 parser (which is used for all of the recent / current models) already has correct code to set the "Serial" string.

src/suunto_vyper_parser.c
```diff
@@ -297,6 +304,17 @@ suunto_vyper_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsi
 				*((dc_divemode_t *) value) = DC_DIVEMODE_OC;
 			}
 			break;
+		case DC_FIELD_STRING:
+			switch(flags) {
+			case 0: /* serial */
+				string->desc = "Serial";
+				snprintf(buf, BUFLEN, "%u", parser->serial);
```
torvalds, line 311: Hmm. I think Suunto serial numbers are actually fixed-length, so they _could_ start with a zero, and this will drop them.

I doubt it matters in practice, and they have some meaning that indicates manufacturing year and month, so it is possible that zero never happens.

I'll merge this as-is, but it's a classic case of "printing it as a number may lose information".

dirkhh: I looked at that trying to find instances of people talking about serial numbers for these older dive computers... I found a hand full of references - and none started with a zero... so... 🤷🏼‍♂️

PR #30, opened 2021-10-31
find lib hidapi on modern openSUSE

dirkhh:
There are now two flavors of hidapi and you need to use a different name
to get one.

Autoconf syntax is... insane.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -126,6 +126,12 @@ AS_IF([test "x$with_hidapi" != "xno"], [
 	AS_IF([test "x$have_hidapi" = "xyes"], [
 		AC_DEFINE([HAVE_HIDAPI], [1], [hidapi library])
 		DEPENDENCIES="$DEPENDENCIES hidapi"
+	], [
+	PKG_CHECK_MODULES([HIDAPI], [hidapi-libusb], [have_hidapi=yes], [have_hidapi=no])
+	AS_IF([test "x$have_hidapi" = "xyes"], [
+		AC_DEFINE([HAVE_HIDAPI], [1], [hidapi library])
+		DEPENDENCIES="$DEPENDENCIES hidapi"
+	])
 	])
 ])
 


jefdriesen:
This shouldn't be necessary. Those two variants of hidapi have always existed on linux (certainly on the debian based distros). The fact that we don't check for the variants explicitly, means hidapi won't be found and we automatically fallback to using libusb directly instead of the hidapi-libusb wrapper. It has always been like that, so I wonder why you want to change this now?

See commit eea02126a41edc42324c5dcaa628ea4d1394441a for details.

dirkhh:
I was working on the fingerprint saving thing and noticed that I couldn't connect to the G2 that Linus loaned me... Forcing it to build against the hidapi-libusb caused it to work.

I also have two random other reports that a G2 and an Eon Steel aren't connecting on openSUSE.

I would prefer if you were right, though 😁

I'll play with it some more

torvalds:
I personally detest autoconfig with a passion, and find it to be the worst possible choice unless you want to support some very odd random legacy Unix platforms. The syntax is obscure.

But I do find other projects that basically do "on Linux, pick 'hidapi-libusb'".  It's not clear why (and it clearly isn't needed on Fedora).

dirkhh:
Yeah, that's why I mention openSUSE in the commit message -- oh, I cleaned that up and don't anymore. Oops.

Fedora doesn't need this. Ubuntu doesn't need this. But for some reason openSUSE appears to. And I don't understand it. I'm planning to set up a clean openSUSE build environment (to make sure it's not something that I messed up) on physical hardware (to make sure that it's not some weird VM issue that somehow doesn't affect the Fedora VM)... basically trying to remove variables.

I assume, Jef, that you never test on openSUSE?

dirkhh:
So here's what I learned so far.
When building with just libusb on openSUSE libdivecomputer `configure` command reasonably says `HIDAPI: NO`.
So far so good.
But then when I try to connect to the G2 from my VM, that fails. Just gets an error opening the device.
Yet, creating a new openSUSE installation on an actual machine, building plain Subsurface 5.0.4 (again with 'HIDAPI: NO'), I **can** successfully download from the G2. So... hmm... you were right Jef, we clearly don't need this.
But I'm really confused what's hosed in that VM to get me this weird result.
Thanks for stopping me from adding something pointless to the code.

Next step is getting back to the two other people with trouble on openSUSE. I think I'll start by asking them if this is in a VM by any chance...

PR #31, opened 2021-11-17
Garmin: Fix gas change event parsing

michaelandreen:
The gas mixes are numbered from 0 and up and the gas change event uses
this number directly, so no need to subtract one.

This is the case for Garmin Mk2i, I don't know if it was different for Mk1.

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -217,7 +217,7 @@ static void garmin_event(struct garmin_parser_t *garmin,
 		return;
 
 	case 57:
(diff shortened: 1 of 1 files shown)


michaelandreen:
What's the status on getting merged @dirkhh ?

dirkhh:
I simply spaced out on it. That's my fault entirely.
Since Linus just opened the Linux merge window, I'll do this myself :)

michaelandreen:
Just wanted to check so it wasn't forgotten for the next subsurface release. =)

PR #32, opened 2022-04-03
OSTC: Use deco model instead of dive mode for deco model information

michaelandreen:
In the most recent merge deco model (byte 79) was swapped for divemode
(byte 82). This caused all OC dives to be classified as the older
algorithm without gradient factors.

Signed-off-by: Michael Andreen <michael@andreen.dev>

### DIFF
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -114,6 +114,7 @@ typedef struct hw_ostc_layout_t {
 	unsigned int duration;
 	unsigned int deco_info1;
 	unsigned int deco_info2;
+	unsigned int deco_model;
 	unsigned int divemode;
 } hw_ostc_layout_t;
 
@@ -170,6 +171,7 @@ static const hw_ostc_layout_t hw_ostc_layout_ostc = {
 	47, /* duration */
 	49, /* deco_info1 */
 	50, /* deco_info1 */
+	0, /* deco_model */
 	51, /* divemode */
 };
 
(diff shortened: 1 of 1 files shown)


michaelandreen:
I haven't actually tested this yet. Need to get development environment up and running on a machine with bluetooth, but it's getting late here.

torvalds:
This looks very likely right to me.

I'd absolutely love to actually have somebody *test* it (I don't have an ostc myself), but this is exactly the kind of thing that that merge would have screwed up, particularly with the confusion between "divemode" and "deco_model".

It's _still_confusing with this, with some dive computer models using "divemode" and others "deco_model".  I don't have docs _or_ hardware, thogh, so....

cyberdiamond:
I should be able to test it in a couple of hours, I have both oc and ccr dives on my ostc3

jb2cool:
This seems to fix the issue for me, here are screenshots built against current master (tagged as bad) and with this adjusted file (tagged as good)
good
bad
.

michaelandreen:
> It's _still_confusing with this, with some dive computer models using "divemode" and others "deco_model". I don't have docs _or_ hardware, thogh, so....

In the current models there the dive mode and deco model are split into two separate attributes: https://code.heinrichsweikamp.com/public/hwos_code/file/tip/doc/hwos_interface.pdf

The older models seem to combine both dive mode and deco model into a single attribute though: https://code.heinrichsweikamp.com/public/mk2/file/tip/docu/ostc_interface_description.odt

So this the confusing mix of divemode and deco_model seems to be the reality of the different hardware that the parser supports.

torvalds:
Thanks everybody, merged into the subsurface branch

PR #33, opened 2022-04-24
Garmin: Handle file names in short format

michaelandreen:
When downloading heart rate data from a chest strap the Descent MK2i
uses a more compact format, like C4ND0302.fit. If any such dives have
been downloaded then newer dives in long format won't be detected and
requires downloading all dives and picking the right ones.

Instead parse this short format and ensure that these dives end up in
the correct order.

Format information found here:
https://gist.github.com/waywardone/fa8cb01462790aa8a26fab97477b80e1

Signed-off-by: Michael Andreen <michael@andreen.dev>

### DIFF
--- a/src/garmin.c
+++ b/src/garmin.c
@@ -141,13 +141,53 @@ struct file_list {
 	struct fit_file *array;
 };
 
+static int
+char_to_int(char c)
+{
+	if (c >= '0' && c <= '9')
+		return c - '0';
+	if (c >= 'a' && c <= 'z')
+		return c - 'a' + 10;
+	if (c >= 'A' && c <= 'Z')
+		return c - 'A' + 10;
+	return 0;
+}
+
+/* C4ND0302.fit -> 2022-04-23-13-03-02.fit */
+static void
+parse_short_name(const char *name, char *output)
+{
+	sprintf(output, "%d-%02d-%02d-%02d-%02d-%02d.fit", char_to_int(name[0]) + 2010, // Year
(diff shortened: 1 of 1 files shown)


src/garmin.c
```diff
@@ -141,13 +141,52 @@ struct file_list {
...
+	sprintf(output, "%d-%02d-%02d-%02d-%02d-%02d.fit", char_to_int(name[0]) + 2010, // Year
+		char_to_int(name[1]), // Month
+		char_to_int(name[2]), // Day
+		char_to_int(name[3]), // Hour
+		char_to_int(name[4]) * 10 + char_to_int(name[5]), // Minute
+		char_to_int(name[6]) * 10 + char_to_int(name[7])); // Second
+}
+
+static char a_buffer[FIT_NAME_SIZE];
+static char b_buffer[FIT_NAME_SIZE];
```
torvalds: Please don't add static variables for something like this. 

These could easily just be regular automatic variables on the stack with no downside.

torvalds (changes requested):
Looks reasonable to me, but I have a nit-picky request.

Yeah, we use random static variables in other places, but let's avoid it when not necessary. If you make those name arrays just be automatic variables on the stack, that 'name_cmp()' function remains thread-safe and doesn't need any static storage.

src/garmin.c
michaelandreen: Done

torvalds:
Thanks, merged. I don't have any test-cases myself, but it looks straightforward to me.

PR #34, opened 2022-05-30
Garmin: decode divemode

dirkhh:
The code used sub_sport to figure out if this was a dive in the first
place, but then never reported back the specific dive mode, leading to
all dives being interpreted as open circuit.

Reported-By: Anton van Rosmalen

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -564,7 +564,22 @@ DECLARE_FIELD(DEVICE_INFO, firmware, UINT16)
 }
 
 // SPORT
-DECLARE_FIELD(SPORT, sub_sport, ENUM) { garmin->dive.sub_sport = (ENUM) data; }
+DECLARE_FIELD(SPORT, sub_sport, ENUM) {
+	garmin->dive.sub_sport = (ENUM) data;
+	dc_divemode_t val;
+	switch (data) {
+	case 55: val = DC_DIVEMODE_GAUGE;
+		break;
+	case 56:
+	case 57: val = DC_DIVEMODE_FREEDIVE;
+		break;
+	case 63: val = DC_DIVEMODE_CCR;
+		break;
+	default: val = DC_DIVEMODE_OC;
+	}
+	garmin->cache.DIVEMODE = val;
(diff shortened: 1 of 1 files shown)


src/garmin_parser.c
```diff
@@ -1502,7 +1517,7 @@ garmin_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsigned i
 	case DC_FIELD_ATMOSPHERIC:
 		return DC_STATUS_UNSUPPORTED;
 	case DC_FIELD_DIVEMODE:
-		return DC_STATUS_UNSUPPORTED;
+		return DC_FIELD_VALUE(garmin->cache, value, DIVEMODE);
```
torvalds, line 1520: This looks fine, but that whole "garmin_parser_get_field()" thing should probably just be deleted, and it should use the standard dc_field_get() helper instead, since it uses all the normal field-cache logic.

src/garmin_parser.c
```diff
@@ -564,7 +564,22 @@ DECLARE_FIELD(DEVICE_INFO, firmware, UINT16)
...
+	case 56:
+	case 57: val = DC_DIVEMODE_FREEDIVE;
+		break;
+	case 63: val = DC_DIVEMODE_CCR;
+		break;
+	default: val = DC_DIVEMODE_OC;
+	}
+	garmin->cache.DIVEMODE = val;
+	garmin->cache.initialized |= 1 << DC_FIELD_DIVEMODE;
+}
```
torvalds, line 582: This should use DC_ASSIGN_FIELD(), but it's not wrong as-is.

I have no idea if those case values are right, but I assume you looked them up somewhere and it all looks sane to me.

torvalds (approved):
All looks good to me, although there is room for cleanup in this driver.

src/garmin_parser.c
dirkhh: done... because I had merged this into the DS9 branch (but not yet updated Subsurface side) I figured I could continue in this branch, but GitHub doesn't like this. So there will be a new merge request and in the end a force push to DS9 oh well

src/garmin_parser.c
dirkhh: done

PR #35, opened 2022-05-31
Garmin divemode

dirkhh:
This is the new one with the cleaned up first commit using the field assign macro and the second commit to replace the get_field function...

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -1491,40 +1491,8 @@ garmin_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsigned i
 
 	if (!value)
 		return DC_STATUS_INVALIDARGS;
-	if (type == DC_FIELD_TANK_COUNT)
-		type = DC_FIELD_GASMIX_COUNT;
 
-	/* This whole sequence should be standardized */
-	if (!(garmin->cache.initialized & (1 << type)))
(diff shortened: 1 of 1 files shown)


torvalds (approved):
Thanks, looks good to me.

I suspect the test for a NULL value pointer could be dropped too, or moved to dc_field_get(), but whatever.

dirkhh:
> 
> I suspect the test for a NULL value pointer could be dropped too, or moved to dc_field_get(), but whatever.

I had briefly thought about that but not spent any more time on it... yeah, that's a logical next cleanup.
Ironically there appears to be only one other user of the `dc_field_get()` helper: `oceans_s1_parser.c` which doesn't check the `value` pointer, either...
In the end, I see no reason not to move that value pointer check into `dc_field_get()`

torvalds:
> Ironically there appears to be only one other user of the `dc_field_get()` helper: `oceans_s1_parser.c` which doesn't check the `value` pointer, either...

Yeah, the whole check is probably bogus, nobody else does it in their get_field() implementations.

And yeah, more backends should use the generic dc_field_get(), but the whole field-cache interface came from the time before there was a helper library, and the dive computers that predate it were not really converted to use it.

Like the Suunto EON Steel one - it could/should use it for everything but the DC_FIELD_TANK case, I think.

Not a big deal.

PR #36, opened 2022-06-14
Simplistic attempt to support new Aqualung i200C

dirkhh:
There appear to be newer models in the wild with a different model
number, but labeled as i200C.

this PR is not sufficient to correctly download everything, but it gets at least somewhat useful support for testing.
The one user with whom I have talked shows that most of the dive data look correct, but the dive date is bogus.

But with this change a user can download things at all, so it's definitely an improvement.

### DIFF
--- a/src/descriptor.c
+++ b/src/descriptor.c
@@ -274,6 +274,7 @@ static const dc_descriptor_t g_descriptors[] = {
 	{"Sherwood", "Amphos 2.0",          DC_FAMILY_OCEANIC_ATOM2, 0x4657, DC_TRANSPORT_SERIAL, NULL},
 	{"Sherwood", "Beacon",              DC_FAMILY_OCEANIC_ATOM2, 0x4742, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
 	{"Aqualung", "i470TC",              DC_FAMILY_OCEANIC_ATOM2, 0x4743, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
+	{"Aqualung", "i200Cv2",             DC_FAMILY_OCEANIC_ATOM2, 0x4749, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
 	/* Mares Nemo */
 	{"Mares", "Nemo",         DC_FAMILY_MARES_NEMO, 0, DC_TRANSPORT_SERIAL, NULL},
 	{"Mares", "Nemo Steel",   DC_FAMILY_MARES_NEMO, 0, DC_TRANSPORT_SERIAL, NULL},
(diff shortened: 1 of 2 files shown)


src/descriptor.c
```diff
@@ -274,6 +274,7 @@ static const dc_descriptor_t g_descriptors[] = {
 	{"Sherwood", "Amphos 2.0",          DC_FAMILY_OCEANIC_ATOM2, 0x4657, DC_TRANSPORT_SERIAL, NULL},
 	{"Sherwood", "Beacon",              DC_FAMILY_OCEANIC_ATOM2, 0x4742, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
 	{"Aqualung", "i470TC",              DC_FAMILY_OCEANIC_ATOM2, 0x4743, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
+	{"Aqualung", "i200C",               DC_FAMILY_OCEANIC_ATOM2, 0x4749, DC_TRANSPORT_SERIAL | DC_TRANSPORT_BLE, dc_filter_oceanic},
```
torvalds: Hmm. Now we have two entries with the exact same text for vendor/model.

It's not obvious in the diff, because the previous entry is  about ten lines earlier.

It might work, but I'm not super-happy about  that kind of ambiguity.  There's no way for a person to actually pick one over the other by name - and that's how people pick these things. Vendor and model name is basically the index into this descriptor array, and indexes should be unique. No?

So i think that "i200C" string should be "i200Cv2" or something.

That said, I suspect it ends up working in practice, with then the atom2_parser code just doing the right thing because it sees the I200CV2 model number and parses things ok in the end.

Hmm?

dirkhh: GAH. I did that in the other part of the support (in Subsurface), but for some reason forgot to do that here.
Good catch. Will fix.
(and yes, the parser shouldn't care about this as it doesn't look at that string)

torvalds:
Anyway, I'm on with this change regardless, so if this has been tested and people prefer the non-unique index, I'm certainly ok with you just merging it.

But if you decide you agree, I'd suggest (a) changing that string to "i200C v2" and (b) moving the i200Cv2 line to be next to the i200C one (despite the numbers not being ordered).

No really strong opinions on my side.

dirkhh:
> But if you decide you agree, I'd suggest (a) changing that string to "i200C v2" and (b) moving the i200Cv2 line to be next to the i200C one (despite the numbers not being ordered).

I had looked at that, but for all the other ones that are just newer versions we did NOT do that. The numbers are ordered, so I decided to stick with this.

I'll merge this once I update the corresponding PR on the Subsurface side to the new SHA for libdc

src/descriptor.c
torvalds: Note that while the parser doesn't look at the string, it *does* look at the model number you pass in.

And I think subsurface will get the model number from the descriptor, and look it up by vendor name and model number, so subsurface will presumably always pick the "v1" model number.

And then libdivecomputer may or may not warn about model numbers having changed.

Or maybe libdivecomputer doesn't end up caring at all. It just worries me a bit.

dirkhh: I think it's just the right thing to do. So I force pushed an updated version that uses i200Cv2 consistently.

In unrelated news... Shearwater listened to our complaints about the "Petrel 2" identifying as "Petrel" on BLE... the new "Perdix 2" and "Petrel 3" use those names on BLE... :)

PR #37, opened 2022-06-14
add Shearwater Perdix 2 and Petrel 3

dirkhh:
Also update and fix the hardware IDs based on the latest information
from Shearwater.
Two of the hardware IDs that we supported before are no listed in the
current docs. In an abundance of caution I'll leave them in the code.
No harm done if they don't exist in real life.

I have tested this with a Perdix 2 that came with the typical one test dive on it. It downloads just fine.
While I got access to the latest hardware IDs, I don't have the updated firmware info, yet, so I expect that there will be a few more changes to add new features - but for now this should get people started.


torvalds (approved):
This obviously looks great.

dirkhh:
I'll wait for the builds (they are triggered in the Subsurface pull request https://github.com/subsurface/subsurface/pull/3468 ) and if they look good, I'll merge this. If I feel super motivated I might even fix the typo in the commit message 🤪

dirkhh:
merged locally with the fixed commit message

PR #39, opened 2022-10-06
Garmin: attempt to parse big endian .FIT files

dirkhh:
The only examples I have seen so far are small incomplete files that don't contain the dive profile, so I'm unsure if this is sufficient (and why there are big endian files in the first place).

This doesn't appear to (and shouldn't) break anything and seems to work ok with the sample files. I think it's worth merging, but could be easily convinced to just ignore it.

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -139,6 +139,7 @@ typedef struct garmin_parser_t {
 	} gps;
 
 	struct dc_field_cache cache;
+	unsigned char is_big_endian; // instead of bool
 } garmin_parser_t;
 
 typedef int (*garmin_data_cb_t)(unsigned char type, const unsigned char *data, int len, void *user);
@@ -390,6 +391,19 @@ struct field_desc {
 	void (*parse)(struct garmin_parser_t *, unsigned char base_type, const unsigned char *data);
 };
 
+static inline int base_type_is_integer(unsigned char base_type)
+{
+	return !memcmp(base_type_info[base_type].type_name + 1, "INT", 3);
+}
+
+static inline unsigned int array_uint_endian(const unsigned char *p, unsigned int type_size, unsigned char bigendian)
+{
+	if (bigendian)
+		return array_uint_be(p, type_size);
+	else
+		return array_uint_le(p, type_size);
+}
+
 #define DECLARE_FIELD(msg, name, type) __DECLARE_FIELD(msg##_##name, type)
 #define __DECLARE_FIELD(name, type) \
 	static void parse_##name(struct garmin_parser_t *, const type); \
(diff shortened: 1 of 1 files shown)


src/garmin_parser.c
```diff
@@ -397,10 +398,19 @@ struct field_desc {
 	{ \
 		if (strcmp(#type, base_type_info[base_type].type_name)) \
 			fprintf(stderr, "%s: %s should be %s\n", #name, #type, base_type_info[base_type].type_name); \
-		type val = *(type *)p; \
+		type val; \
+		char is_endian_int = base_type_info[base_type].type_name[0] == 'S' || base_type_info[base_type].type_name[0] == 'U'; \
```
torvalds: Ugh. This is nasty. The 'S' will also trigger on 'STRING', and then we'll rely on that "type_size > 1" check to not do anything.

So it works, but..

I'd rather you do something like

     !memcmp(base_type_info[base_type].type_name+1, "INT")

but preferably in a helper inline function like "base_type_is_integer()" so that it would be a lot more legible.

src/garmin_parser.c
```diff
@@ -397,10 +398,19 @@ struct field_desc {
...
+		if (base_type_info[base_type].type_size > 1 && is_endian_int) { \
+			if (g->is_big_endian) \
+				val = array_uint_be(p ,base_type_info[base_type].type_size); \
+			else \
+				val = array_uint_le(p ,base_type_info[base_type].type_size); \
+		} else { \
+			val = *(type *)p; \
+		} \
```
torvalds: Can we add a "array_uint_endian(ptr, size, bigendian)" helper, and just use it here instead of this ugly thing?

Please?

src/garmin_parser.c
```diff
@@ -1217,17 +1227,18 @@ static int traverse_definition(struct garmin_parser_t *garmin,
 	struct type_desc *desc = garmin->type_desc + type;
 	int fields, devfields, len;
 
-	msg = array_uint16_le(data+2);
+	msg = data[1] ? array_uint16_be(data+2) : array_uint16_le(data+2);
```
torvalds: .. and then use the helper here too.

torvalds (approved):
The patch is ok and you can merge it. 

I have some stylistic comments in case you want to go the extra mile, but up to you.

src/garmin_parser.c
dirkhh: yeah, I like that better. I was struggling with the variable name - because that is what really shows how dumb this is (`STRING` isn't an endian int). I had noticed the issue, but figured with the `type_size > 1` check I was ok. But you are right. It's dumb. I'll fix it and move the whole thing into a helper.

src/garmin_parser.c
dirkhh: sure... there is lots of room for making this prettier.

torvalds:
That last version looks fine to me. But that's just from a quick scan-through, so what do I know...

dirkhh:
I tested it on both real data from my Mk1 and Mk2 dive computers as well as the "short" big endian files from Larry.
So I'm somewhat optimistic that this might at least not break things.
I'll merge it.

PR #40, opened 2022-11-23
Added parsing of the CCR setpoint information for Garmin Descent computers.

mikeller:
This is adding support for parsing the setpoint information in logfiles downloaded from Garmin Descent devices.

Screenshot from 2022-11-24 11-25-25

The Garmin devices do not have support for ppO2 sensor input, so they only work in 'fixed setpoint' mode for CCR dives, and dive data records do not contain actual ppO2 values. The ppO2 values are retrofitted to the dive data based on 'setpoint change' events reported by the device. With this change CCR dives downloaded from a Garmin device are correctly classified as CCR dive, and the calculated ceiling / tissue loading graphs are accurate and match the deco stops reported by the device. Before this change, CCR dives were classified as open circuit dives, often resulting in a massively overstated calculated ceiling.

This has been tested for logs with only automated setpoint changes - more test dives are needed to reverse engineer the log file format for manual setpoint changes, as the setpoint fields are not documented in Garmin's documentation for the FIT file format.

Signed-off-by: Michael Keller <github@ike.ch>

PR #41, opened 2022-12-08
Added detection and status messages for bailout to OC / switch back to CC for CCR dives.

mikeller:
This might need a bit of discussion.
As a CCR diver I would like to see the ability to properly track the 'on loop' / 'bailed out to open circuit' status directly in Subsurface, because this is an important (or even the most important) bit of status information during a CCR dive. This should extend to the use of the correct ppO2 / gas mix in the deco ceiling / tissue model calculation. My idea for how to do this would be to track the 'type' ('diluent' / 'OC bailout') for every gasmix / tank that is reported by the dive computer. Most CCR capable dive computers that I am familiar with require the user to enter two different gas lists for diluent and bailout, so this should work with the existing libdivecomputer API for these.
Unfortunately I think making this change in Subsurface will require a bit of work, as the libdivecomputer field capable of tracking the 'type' of a gas or tank (`cache->tankinfo[]`) does not seem to be consumed at all in Subsurface. So this pull request is just providing a prerequisite for the change in Subsurface by populating `tankinfo[]`. In addition to this it also triggers a message on every switch from CC to OC and back, at least giving a visual indication of these diver triggered events. The messages can probably be removed from libdivecomputer again once 'loop status' tracking has been added to Subsurface.

image

Also included is a fix of the tab expansion mess that I created in https://github.com/subsurface/libdc/pull/40. Apologies for this, I've switched to using a custom `.vimrc` for this project now.

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -71,7 +71,7 @@ struct record_data {
 	double ceiling;
 
 	// RECORD_GASMIX
-	int index, gas_status;
+	int index, gas_status, gas_type;
 	dc_gasmix_t gasmix;
 
 	// RECORD_EVENT
@@ -127,7 +127,8 @@ typedef struct garmin_parser_t {
 		struct garmin_sensor sensor[MAX_SENSORS];
 		unsigned int setpoint_low_cbar, setpoint_high_cbar;
 		unsigned int setpoint_low_switch_depth_mm, setpoint_high_switch_depth_mm;
-        unsigned int setpoint_low_switch_mode, setpoint_high_switch_mode;
+		unsigned int setpoint_low_switch_mode, setpoint_high_switch_mode;
+		dc_tankinfo_t *current_tankinfo;
 	} dive;
 
 	// I count nine (!) different GPS fields Hmm.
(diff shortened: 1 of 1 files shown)


mikeller:
I hate to ask, but is it possible to get some traction on this one?
I am trying to improve CCR related gas logging for other dive computer models as well, and getting this (and #43) reviewed will help to clarify that the direction this is going is right.

dirkhh:
I poked Linus - he doesn't always see GitHub notifications.

torvalds:
Hmm. The commit message is a bit of a mess on that first commit, with overlong lines (that's a 702-character line!)

We had that earlier with the "Added parsing of the CCR setpoint information for Garmin Descent computers." that I merged as part of "added_garmin_ccr_setpoint_info", and I dread getting more of these.

But the changes look sane per se.

mikeller:
Apologies, I am new to the approach to add a detailed description of the changes to the commit message, instead of adding it to the pull request, and did not take care to rewrap my commit messages after editing. Fixed now.

torvalds:
Cherry-picked the one remainming commit in the updated pull.

PR #43, opened 2023-01-17
Refactoring: Push field cache access down into base class.

mikeller:
Pushed access to cached fields down into `field-cache.c` from parsers that use the field cache.
As a result, all cached fields will be available to Subsurface in computer models that use the field cache. In particular this means that, together with https://github.com/subsurface/libdc/pull/41, the proper use type (diluent / OC bailout) of tanks will be shown for the Garmin Descent dive computers (others to follow).

image

### DIFF
--- a/src/deepblu_parser.c
+++ b/src/deepblu_parser.c
@@ -210,33 +210,11 @@ deepblu_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsigned 
 		return DC_STATUS_UNSUPPORTED;
 
 	switch (type) {
-	case DC_FIELD_DIVETIME:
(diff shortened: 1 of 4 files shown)


torvalds:
I have merged this, although I ended up doing so by cherry-picking the commit with edits to both the commit message and to avoid compile warnings from switch statements after cases were removed.

mikeller:
Thanks for your help @torvalds, and apologies for missing the compiler warnings.

torvalds:
Please do check what else is missing from your work, and create a new pull request for that.

I think there was some code to keep the cylinder number and tank info bits in sync.

mikeller:
@torvalds: Checked, it's all there, and it's working as intended.

PR #44, opened 2023-02-13
HW OSTC4: Fix and improve CCR gas handling.

mikeller:
Fixed a bug for the OSTC4 using the wrong gas after (diluent) gas changes because the diluent gas index is offset by 5 when in CCR mode. Also added proper tracking of gas use type (diluent / OC bailout) during CCR dives, and added explicit events for OC bailout / switch back to closed circuit.
This intentionally only handles OSTC4 data - I do not own any other Heinrichs Weikamp devices, and have got no easy way to figure out if any of their other models with CCR support have the same quirk of remapping diluent gases in the log header. If somebody owns such a device and can test for me I am happy to make the changes for other devices.

image

### DIFF
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -1057,7 +1057,7 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				return DC_STATUS_DATAFORMAT;
 			}
 			unsigned int idx = data[offset];
-			if (parser->model == OSTC4 && ccr && idx > parser->nfixed && idx <= parser->nfixed + OSTC4_CC_DILUENT_GAS_OFFSET) {
+			if (idx > parser->nfixed && idx <= parser->nfixed + OSTC4_CC_DILUENT_GAS_OFFSET) {
 				// OSTC4 reports gas changes to another diluent with an offset
 				idx -= OSTC4_CC_DILUENT_GAS_OFFSET;
 			}
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -140,6 +140,7 @@ typedef struct hw_ostc_parser_t {
 	unsigned int initial_setpoint;
 	unsigned int initial_cns;
 	hw_ostc_gasmix_t gasmix[NGASMIXES];
+	unsigned int current_divemode_ccr;
 } hw_ostc_parser_t;
 
 static dc_status_t hw_ostc_parser_set_data (dc_parser_t *abstract, const unsigned char *data, unsigned int size);
@@ -801,6 +802,27 @@ hw_ostc_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsigned 
 }
 
 
+static void hw_ostc_notify_bailout(hw_ostc_parser_t *parser, const unsigned char *data, unsigned int index, dc_sample_callback_t callback, void *userdata)
+{
+	if (parser->current_divemode_ccr != parser->gasmix[index].diluent) {
+		dc_sample_value_t sample = {
+			.event.type = SAMPLE_EVENT_STRING,
+			.event.flags = SAMPLE_FLAGS_SEVERITY_INFO,
+		};
+		if (parser->gasmix[index].diluent) {
+			sample.event.name = "Switched to closed circuit";
+		} else {
+			sample.event.name = "Switched to open circuit bailout";
+		}
+
+		if (callback) {
+			callback(DC_SAMPLE_EVENT, sample, userdata);
+		}
+
+		parser->current_divemode_ccr = parser->gasmix[index].diluent;
+	}
+}
+
 static dc_status_t
 hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t callback, void *userdata)
 {
@@ -915,6 +937,7 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 
 	// Get the CCR mode.
 	unsigned int ccr = hw_ostc_is_ccr (divemode, version);
+	parser->current_divemode_ccr = ccr;
 
 	unsigned int time = 0;
 	unsigned int nsamples = 0;
@@ -1046,6 +1069,9 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 
 			sample.gasmix = idx;
 			if (callback) callback (DC_SAMPLE_GASMIX, sample, userdata);
+
+			hw_ostc_notify_bailout(parser, data, idx, callback, userdata);
+
 			offset += 2;
 			length -= 2;
 		}
@@ -1069,6 +1095,9 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 			sample.gasmix = idx;
 			if (callback) callback (DC_SAMPLE_GASMIX, sample, userdata);
 			tank = idx;
+
+			hw_ostc_notify_bailout(parser, data, idx, callback, userdata);
+
 			offset++;
 			length--;
 		}
@@ -1111,6 +1140,9 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 
 				sample.gasmix = idx;
 				if (callback) callback (DC_SAMPLE_GASMIX, sample, userdata);
+
+				hw_ostc_notify_bailout(parser, data, idx, callback, userdata);
+
 				offset += 2;
 				length -= 2;
 			}
@@ -1244,6 +1276,9 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 
 				sample.gasmix = idx;
 				if (callback) callback (DC_SAMPLE_GASMIX, sample, userdata);
+
+				hw_ostc_notify_bailout(parser, data, idx, callback, userdata);
+
 				offset += 2;
 				length -= 2;
 			}
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -39,6 +39,7 @@
 
 #define MAXCONFIG 7
 #define NGASMIXES 15
+#define OSTC4_CC_DILUENT_GAS_OFFSET 5
 
 #define UNDEFINED 0xFFFFFFFF
 
@@ -1021,8 +1022,15 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				return DC_STATUS_DATAFORMAT;
 			}
 			unsigned int o2 = data[offset];
+			unsigned int diluent;
+			if (parser->model == OSTC4) {
+				// all manually added gas mixes on OSTC4 are OC gases
+				diluent = 0;
+			} else {
+				diluent = ccr;
+			}
 			unsigned int he = data[offset + 1];
-			unsigned int idx = hw_ostc_find_gasmix (parser, o2, he, ccr, MANUAL);
+			unsigned int idx = hw_ostc_find_gasmix (parser, o2, he, diluent, MANUAL);
 			if (idx >= parser->ngasmixes) {
 				if (idx >= NGASMIXES) {
 					ERROR (abstract->context, "Maximum number of gas mixes reached.");
@@ -1032,7 +1040,7 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				parser->gasmix[idx].helium = he;
 				parser->gasmix[idx].type = 0;
 				parser->gasmix[idx].enabled = 1;
-				parser->gasmix[idx].diluent = ccr;
+				parser->gasmix[idx].diluent = diluent;
 				parser->ngasmixes = idx + 1;
 			}
 
@@ -1049,9 +1057,9 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				return DC_STATUS_DATAFORMAT;
 			}
 			unsigned int idx = data[offset];
-			if (parser->model == OSTC4 && ccr && idx > parser->nfixed) {
-				// Fix the OSTC4 diluent index.
-				idx -= parser->nfixed;
+			if (parser->model == OSTC4 && ccr && idx > parser->nfixed && idx <= parser->nfixed + OSTC4_CC_DILUENT_GAS_OFFSET) {
+				// OSTC4 reports gas changes to another diluent with an offset
+				idx -= OSTC4_CC_DILUENT_GAS_OFFSET;
 			}
 			if (idx < 1 || idx > parser->nfixed) {
 				ERROR(abstract->context, "Invalid gas mix (%u).", idx);
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -573,6 +573,7 @@ hw_ostc_parser_get_field (dc_parser_t *abstract, dc_field_type_t type, unsigned 
 				return DC_STATUS_UNSUPPORTED;
 			*((double *) value) = array_uint16_le (data + layout->avgdepth) / 100.0;
 			break;
+		case DC_FIELD_TANK_COUNT:
 		case DC_FIELD_GASMIX_COUNT:
 			*((unsigned int *) value) = parser->ngasmixes;
 			break;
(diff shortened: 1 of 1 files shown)


src/hw_ostc_parser.c
```diff
@@ -1049,6 +1080,10 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				return DC_STATUS_DATAFORMAT;
 			}
 			unsigned int idx = data[offset];
+			if (version == 0x24 && data[layout->divemode] == OSTC3_CC) {
+				// OSTC4 reports gas changes to another diluent with an offset
+				idx -= CC_DILUENT_GAS_OFFSET;
```
torvalds: Is this offset actually guaranteed to be there? What if this isn't a diluent gas at all?

IOW, maybe this all shouldn't be conditional on "version 0x24 and CC", but on 'idx' having a large value?

torvalds (changes requested):
Patch looks sane to me apart from that one "subtract 5" conditional being really odd.

Maybe it's correct, but I think it needs more explanation.

src/hw_ostc_parser.c
mikeller: From looking at the [source](https://code.heinrichsweikamp.com/public/ostc4/file/tip) for the firmware currently released for the OSTC4, this offset is applied on the OSTC4 _for every dive in CCR mode_. The offset is _not_ applied on the OSTC4 when it is in OC mode - so this check is definitely needed.
As for the "version is `0x24` check, the background for this is that both OSTC4 and a  number of other Heinrichs Weikamp dive computers are both using the 'ostc3' 'dialect' of the hw_ostc protocol. The differentiation between the 'sub-dialects' is made by the `version` byte, with `0x24` standing for OSTC4. (I think using defines for `0x23` and `0x24` would greatly simplify reading this code.)
These other dive computers use a [completely different firmware](https://code.heinrichsweikamp.com/public/hwos_code/file/tip/src), and from talking to @jefdriesen this firmware does not have this offset for gas changes - I cannot test this.
Also from looking at the OSTC4 source, only changes to a diluent gas are reported through a `0x20` gaschange event - changes to an OC bailout gas are always reported as a `0x0100` bailout event.
As for generally applying an offset if the reported index is in between 5 and 10 I am confident that this would work for the current firmware for OSTC4, but not sure if this will work for the other firmware using the same code path - there might be cases where it legitimately returns gas mixes in this range.

Also note that I've pushed another update - I realised that during the rebase I had accidentally removed the code pushing the gas use type back into Subsurface.

src/hw_ostc_parser.c
```diff
@@ -1032,12 +1070,20 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				parser->gasmix[idx].helium = he;
 				parser->gasmix[idx].type = 0;
 				parser->gasmix[idx].enabled = 1;
-				parser->gasmix[idx].diluent = ccr;
+				if (version == 0x24) {
```
mikeller: @torvalds: This is another case where I was able to confirm for OSTC4 (`0x24`) that the current implementation is broken - but not for the rest of the dive computers using this code path.

mikeller:
Added a fix for manual gas changes.

src/hw_ostc_parser.c
jefdriesen: This check is wrong. You can't use the version to identify the OSCT4, because version 0x24 is also used by the hwOS devices. So this change will certainly cause trouble for those models. For the OSTC4 you need to check for parser->model == OSTC4.

src/hw_ostc_parser.c
```diff
@@ -1021,8 +1059,15 @@ hw_ostc_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callback_t call
 				return DC_STATUS_DATAFORMAT;
 			}
 			unsigned int o2 = data[offset];
+			unsigned int diluent;
+			if (version == 0x24) {
+				// all manually added gas mixes on OSTC4 are OC gases
+				diluent = 0;
```
jefdriesen, line 1065: Besides the fact that the check for version 0x24 is wrong (see the other comment), the OSTC4 doesn't seem to produce manual gas change events in CC mode. So unless I missed something, I don't think this change will have any effect in practice. Did you encounter a problem here? Note that if the OSTC4 firmware ever gains support for manual gas change events, this code will be wrong.

In the OSTC4 code, the only place where the manualGasSet event is set is in the setActualGas_ExtraGas() function, and that is only called from OnAction_DM_Mix():
https://code.heinrichsweikamp.com/public/ostc4/file/tip/Discovery/Src/data_central.c#l538

jefdriesen (changes requested):

src/hw_ostc_parser.c
mikeller: > the OSTC4 doesn't seem to produce manual gas change events in CC mode.

I think the `/* only for OC */` comment on `OnAction_DM_Mix()` is a bit misleading - what it means is that the `Extra Gas` change always results in the new gas being used as OC / bailout - it is not possible to switch to an extra diluent during the dive:

https://code.heinrichsweikamp.com/public/ostc4/file/tip/Discovery/Src/tMenuEditGasOC.c#l348

And yes, if the ability to switch to extra diluent gases during the dive is ever added for OSTC4 then this will have to be reworked. But I am not concerned about this, as currently the log format does not have a way to indicate that an extra gas being switched to is a diluent, so a protocol change (likely to break backwards compatibility) will have to happen for this.

mikeller:
@jefdriesen:

In response to https://github.com/subsurface/libdc/pull/44#discussion_r1126128408:

Thanks for letting me know. Is there a published list of of version information for Heinrichs Weikamp?

Using `parser->model == OSTC4` is somewhat problematic in Subsurface, as `parser->model` is derived from user entered information, so this will fail in unpredictable ways, returning garbage data and potentially resulting in an application crash if the user selects the wrong device.

I'll be looking into asserting the device type in Subsurface before trying to download logs, similar to what is done https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/hw_ostc3.c#L1612 to make this a bit more robust.

jefdriesen:
The latest hwOS documentation is available in the public git repository: https://code.heinrichsweikamp.com/public/hwos_code/file/tip/doc
There is no separate documentation for the OSTC4 because it's supposed to be 100% compatible, and except for a few quirks here and there, that's also the case.

The `parser->model` is indeed passed by the caller when creating the parser. But that does not necessary mean it's a value controlled by the end-user. During a download, the user needs to select a model (which may be any model from the correct family), but the actual model is also auto-detected during the communication. If the application does this correctly (which is always the case when using the `dc_parser_new()` function), then that's the model number that will be passed to the parser, and not whatever model the end-user did select.

There are ways to pass the wrong model number. For example when trying to parse data 'offline' (e.g. not immediately after the download). But that should certainly not result in a crash. If it does, please report because that's a serious bug that should be fixed. It may result in incorrect data, but that's something different.

As far as I know, there is no way to distinguish the OSTC4 from the other hwOS devices based only on the dive data. That's by design because it's supposed to be 100% compatible. In those few cases where that's not true, it would be nice to have such info, but unfortunately that's not the case.

src/hw_ostc_parser.c
jefdriesen: There is indeed no explicit indication whether a manual gas is an OC mix or diluent. That info is obtained indirectly from the dive mode. The assumption here is that all manual gas mixes are OC gas mixes when in OC mode and diluents when in CC mode. But that's the code you removed.

Note that bailout to OC has it's own event, so there we do know we're switching to an OC gas mix.

mikeller:
> The latest hwOS documentation is available in the public git repository: [code.heinrichsweikamp.com/public/hwos_code/file/tip/doc](https://code.heinrichsweikamp.com/public/hwos_code/file/tip/doc) There is no separate documentation for the OSTC4 because it's supposed to be 100% compatible, and except for a few quirks here and there, that's also the case.

Thanks. https://code.heinrichsweikamp.com/public/hwos_code/file/tip/doc/hwos_interface.pdf is what I was looking for, and wasn't able to find in the repository for the OSTC4 source / documentation.

> but the actual model is also auto-detected during the communication. If the application does this correctly (which is always the case when using the `dc_parser_new()` function)

Can you please elaborate a bit on where this is happening - all I can see is that whatever user entered value for the model is submitted in `device` in the call to `dc_parser_new()` is then set as the model in `hw_ostc_parser_create()`:

https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/parser.c#L220-L230

https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/parser.c#L77-L217

> As far as I know, there is no way to distinguish the OSTC4 from the other hwOS devices based only on the dive data. That's by design because it's supposed to be 100% compatible. In those few cases where that's not true, it would be nice to have such info, but unfortunately that's not the case.

I understand this, and I think having a device independent format is a good thing in general.
Unfortunately, as in the case that we are discussing here, there seem to be little device-specific quirks that mean the format is not 100% device independent.

src/hw_ostc_parser.c
mikeller: > The assumption here is that all manual gas mixes are OC gas mixes when in OC mode and diluents when in CC mode. 

I try to base my code on established facts and not on assumptions, whereever this is possible.
And for the OSTC4 we *know* that a switch to an 'Extra Gas' is always treated as a switch to an *OC* gas - which is what I am modelling in the code that I am proposing.
Maybe we should try to get certainty on what is happening in the 'Extra Gas' case for the non-OSTC4 dive computers using this log format as well, to be able to base all of this code on established facts?
I am not familiar with the code base for these (or indeed what models and potential firmware versions exist), and I don't have the hardware to test this.
Or we could ask the manufacturer.

jefdriesen:
The `device->devinfo.model` is not user entered info! For most device backends, the exact model selected by the user doesn't even get passed to the backend at all. In the `dc_device_open()` function, the family type is always used to select the correct backend, but the model number is typically not used. The `hw_ostc3_device_open()` function does not have a parameter for it:

https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/device.c#L186

The DC_EVENT_DEVINFO event is emitted in the backend, with info that is obtained from the dive computer (`device->hardware`):

https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/hw_ostc3.c#L746-L755

Next, in the `device_event_emit()` function the event data is saved in the device struct:

https://github.com/subsurface/libdc/blob/8d3271e586cdef790197effb3985c04f35827d00/src/device.c#L483

And finally the parser retrieves the model number from there.

mikeller:
@jefdriesen: Ah, so what you are saying is that the `device` supplied to `dc_parser_new()` **must be** what is returned by `dc_device_open()` and not user entered data for the parser to correctly parse the data?
Apologies, but I did not understand this from your previous comment, and wasn't able to see it from looking at the code.

jefdriesen:
A `dc_device_t` handle represents a connection to the dive computer. The only way to create one is with the `dc_device_open()` function. Thus no user entered data here.

Note that there is also a `dc_parser_new2()` function, which does not require a device handle. This function exist to support "offline" parsing for use-cases where there is no device handle available (anymore). For example, after the download you can save the raw dive data to disk, and parse it another time when the dive computer is no longer connected. This function accepts a model number through the `dc_descriptor_t` parameter, and here you could pass the wrong model. It's the responsibility of the caller to also pass the correct model.

mikeller:
Ah, I see. Unfortunately this didn't become clear to me from looking at the code, probably because `dc_device_t` is a data structure containing data, and not a [handle](https://en.wikipedia.org/wiki/Handle_(computing)). Maybe it's handle-like character should be hinted at with a name like `dc_device_handle_t`?

jefdriesen:
The `dc_device_t` (and many other similar types in libdivecomputer) is a typedef to an [opaque](https://en.wikipedia.org/wiki/Opaque_pointer) struct. That means you can declare a pointer to such a structure, but not dereference it because the actual definition is private. Thus the pointer acts as a handle.

mikeller:
@torvalds: Is there anything more that needs to be done for this - from my point of view this is as close as we can get with making these conditions generic, taking into account that we don't know how what data is returned by the non-OSTC4 computers using this protocol.

torvalds:
> @torvalds: Is there anything more that needs to be done for this 

This still does that 

+			if (version == 0x24 && data[layout->divemode] == OSTC3_CC) {
+				// OSTC4 reports gas changes to another diluent with an offset
+				idx -= CC_DILUENT_GAS_OFFSET;
+			}

that I absolutely hated. 

What is wrong with just using

	// OSTC4 reports gas changes to another diluent with an offset
	if (idx > parser->nfixed)
		idx -= CC_DILUENT_GAS_OFFSET;

and then at that point continuing on with the existing

	if (idx < 1 || idx > parser->nfixed) {
		ERROR(abstract->context, "Invalid gas mix.");
		return DC_STATUS_DATAFORMAT;
	}

because the whole "based on device version and divemode" seems quite odd to me. The parsing isn't _supposed_ to differ between different OSTC computers.

I suspect that what the whole CC_DILUENT_GAS_OFFSET is meant to be is simply a flag to say "this is a diluent gas change" as opposed to an OC gas change. The DC_SAMPLE_GASMIX sample just doesn't have that kind of field, so libdivecomputer essentially just drops it.

mikeller:
> > @torvalds: Is there anything more that needs to be done for this
> 
> This still does that
> 
> * ```
>     	if (version == 0x24 && data[layout->divemode] == OSTC3_CC) {
>   ```
> * ```
>     		// OSTC4 reports gas changes to another diluent with an offset
>   ```
> * ```
>     		idx -= CC_DILUENT_GAS_OFFSET;
>   ```
> * ```
>     	}
>   ```
> What is wrong with just using
> 
> ```
> // OSTC4 reports gas changes to another diluent with an offset
> if (idx > parser->nfixed)
> 	idx -= CC_DILUENT_GAS_OFFSET;


I think you need to refresh - I did push a change to address this:

https://github.com/subsurface/libdc/pull/44/files#diff-4004197d7170e5df533d87b83e1b00d179063d881cc7b7de3c784298ad137530R1100

I've added a second condition `idx <= parser->nfixed + OSTC4_CC_DILUENT_GAS_OFFSET` to the check to address cases where the index is outside of the range that we know is returned by OSTC4 for diluent changes - this is really just cosmetic to avoid returning invalid information in the check further down: `ERROR(abstract->context, "Invalid gas mix (%u).", idx)`.

> because the whole "based on device version and divemode" seems quite odd to me. The parsing isn't _supposed_ to differ between different OSTC computers.

I know. But unfortunately at this point we _know_ that the data format returned by different dive computer models _is_ different.

> I suspect that what the whole CC_DILUENT_GAS_OFFSET is meant to be is simply a flag to say "this is a diluent gas change" as opposed to an OC gas change. The DC_SAMPLE_GASMIX sample just doesn't have that kind of field, so libdivecomputer essentially just drops it.

(Only speaking for the OSTC4 source code here:) Pretty much - except that the OSTC4 will never send a gasmix index <= 5 (1 based) for a change to an open circuit gas - these are all sent as changes to a 'custom gas'.

torvalds:
On Tue, Mar 14, 2023 at 5:50 PM Michael Keller ***@***.***> wrote:
>
> I think you need to refresh - I did push a change to address this:

Ahh. I did check the current git tree, but I was walking the
individual commits, and saw that broken case.

I didn't then notice that you fixed it up in the next one.

                Linus

torvalds:
So in order to make things easier to merge with Jef's upstream, I've taken his workaround for now.

It conflicts with yours, but that's not a real problem - the conflict is trivial and I could easily do these kinds of small conflicts in my sleep. I deal with much larger conflicts in the kernel.

But before I take your changes, I would like you to split them up a bit.

In particular, can you do that DC_FIELD_TANK_COUNT and DC_FIELD_TANK part as one commit, and then do the whole gas parsing part as another?

mikeller:
> But before I take your changes, I would like you to split them up a bit.
> 
> In particular, can you do that DC_FIELD_TANK_COUNT and DC_FIELD_TANK part as one commit, and then do the whole gas parsing part as another?

Good point - we will want to get rid of using `dc_tank_t` as a workaround to encode gas use once the libdivecomputer protocol has changed to have gas use in `dc_gasmix_t`. (I'll also happily remove changes like this manually once we get to that point.)

I've also separated out the bailout / switch back notification into a separate commit - we'll probably want to get rid of this once Subsurface can display bailout events for logged dives based on the gas use changing during a gas switch, which is an item on my TODO list.

And lastly I have split the change making Jef's code more generic, instead of only trying to apply the gasmix index offset on OSTC4 and for CCR dives. On reflection I am again not sure if this is right:
- we do know (from reading the OSTC4 source code and testing) that this offset is applied on OSTC4 _only for CCR dives_;
- we do know (from logfiles that Jef has) that this offset is not applied on other dive computers using the same protocol, even on CCR dives.

I fully get your point about the protocol meant to be generic and work across multiple dive computers, but I think this is a case where this doesn't apply, based on a design quirk, or a bug, introduced by the manufacturer.
Not sure what Heinrich Weikamp's take on this is, but it might be difficult for them to fix this in the firmware without breaking backwards compatibility in the protocol, and creating problems for other implementations of logging software that have this coded as a fixed offset applied to all OSTC4 CCR dives.

torvalds:
Merged.

PR #46, opened 2023-02-20
libdivecomputer update

torvalds:
This is my merge of Jef's current work. It builds cleanly on Linux, which is currently all I can say about it.

I think I merged things correctly, and it all *looks* fine, but some of the merge was a bit painful due to Jef adding features near ours. In particular some of the decomode stuff is just silly and pointless, and exists purely because Jef still can't handle string data. 

I'm *not* pushing it to the upstream libdc repo because of lack of testing, but with the merge window opening up tomorrow I also probably won't have much time for doing said testing. Not great. Not that I think that there's anything wrong here, but mistakes can happen.

So it would be interesting if you can test this out with your current subsurface, but even if it tests out fine (I did build the two together, saw no issues), I'd worry about what a github merge does. So caveat emptor.


mikeller:
A bit late to the party, apologies, I had other high priority stuff on for the last few days.

I did import tests with the latest on subsurface/master and libdivecomputer/Subsurface-DS9 with this merged in for Garmin (Descent Mk1), Heinrichs Weikamp (OSTC4), and Shearwater (Petrel 2).

Garmin and Shearwater are fine, but unfortunately with the OSTC4 I got 3 header parsing errors (see below) out of 12 dives. I repeated the same download with latest on subsurface and 6200a7923f1539daa3c6273e2b9f4eaf7b7dbc71 on libdivecomputer (right before this was merged), and all 12 dives were parsed successfully.

I'll have an investigation of this and see if I can figure out what's going on.

```
[15.779945] ERROR: Invalid gas mix (7). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
ERROR: Dive 1: Error parsing the divetime
ERROR: Dive 1: Error parsing the header
[35.018690] ERROR: Invalid gas mix (6). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
ERROR: Dive 11: Error parsing the divetime
ERROR: Dive 11: Error parsing the header
[36.223246] ERROR: Invalid gas mix (6). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
ERROR: Dive 12: Error parsing the divetime
ERROR: Dive 12: Error parsing the header
Finishing download thread: 9 dives downloaded
```

torvalds:
On Tue, Feb 21, 2023 at 3:24 AM Michael Keller ***@***.***> wrote:
>
> A bit late to the party, apologies, I had other high priority stuff on for the last few days.

Well, quite the fast reaction to me merging Jef's upstream, honestly.

> I did import tests with the latest on subsurface/master and
> libdivecomputer/Subsurface-DS9 with this merged in for Garmin (Descent
> Mk1), Heinrichs Weikamp (OSTC4), and Shearwater (Petrel 2).
>
> Garmin and Shearwater are fine, but unfortunately with the OSTC4 I got
> 3 header parsing errors (see below) out of 12 dives. I repeated the
> same download with latest on subsurface and 6200a79 on libdivecomputer
> (right before this was merged), and all 12 dives were parsed
> successfully.

I'm in the middle of the kernel merge window, which is a bit
inconvenient, but this does not look like a libdivecomputer merge
issue.

I was worried about the OSTC conflicts, but was quite careful, and the
only differences in the OSTC parsing code now are our additions for
battery voltage and the string interface.

So it looks like a problem with Jef's upstream, and I've added Jef to the cc.

> I'll have an investigation of this and see if I can figure out what's going on.
>
> [15.779945] ERROR: Invalid gas mix (7). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
> ERROR: Dive 1: Error parsing the divetime
> ERROR: Dive 1: Error parsing the header
> [35.018690] ERROR: Invalid gas mix (6). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
> ERROR: Dive 11: Error parsing the divetime
> ERROR: Dive 11: Error parsing the header
> [36.223246] ERROR: Invalid gas mix (6). [in src/hw_ostc_parser.c:1053 (hw_ostc_parser_samples_foreach)]
> ERROR: Dive 12: Error parsing the divetime
> ERROR: Dive 12: Error parsing the header
> Finishing download thread: 9 dives downloaded

Judging by the error message, I suspect it's commit 328812e ("Limit
the index to the fixed gas mixes").

Can you try if reverting that commit just fixes it?

                 Linus

jefdriesen:
It's a different issue. The hwOS devices support a maximum of 5 fixed gas mixes. More gases can be added during the dive, but those are all manual gases which are handled different. So the index shouldn't exceed the number of fixed gas mixes, and that's what commit 328812e95bfe7c6c9d2a8d36c75144f05c7dc9dc is about.

But it appears that for CCR dives, the OSTC4 behaves a bit different from the other hwOS models and uses index 6 to 10 for the diluents. I've already reported this to Matthias, and I'm currently waiting for feedback. If this is indeed confirmed to be an OSTC4 firmware issue, then the workaround will be to subtract 5 from the index.

mikeller:
@torvalds: I had hoped to get around to testing before it was merged, to get the opportunity to exclude problematic changes before the merge. But alas.

I did a quick check with the changes in 328812e95bfe7c6c9d2a8d36c75144f05c7dc9dc manually backed out, and this indeed fixes it. This means that this change breaks support for OSTC4.
I've included a fix for this, and for another place where support for OSTC4 is broken in the now rebased #44.

One thing that I am noticing is that the error messages returned after parsing problems often seem to be misleading:

```
ERROR: Dive 1: Error parsing the divetime
```

From the look of it, libdivecomputer seems to be aborting log parsing and returning `DC_STATUS_DATAFORMAT` for any field access, even if the format inconsistency was discovered in a completely unrelated place.
Some options to improve things would be:
- remove the detailed error messages from Subsurface as they are misleading;
- have libdivecomputer return a more generic error message like `DC_STATUS_PARSING_FAILED` if parsing is aborted, and not try to pinpoint the location of the failure in Subsurface;
- make libdivecomputer more resilient and skip fields / samples that cannot be correctly parsed, but try to continue parsing the rest of the information.

torvalds:
> * make libdivecomputer more resilient and skip fields / samples that cannot be correctly parsed, but try to continue parsing the rest of the information.

At least in this case, that's probably the right thing to do.

It's kind of annoying how a small error in libdivecomputer due to an incomplete understanding of the rules then made the whole parse fail. It should have reported the failure, but then just continued. Possibly with some logic like "after lots of warnings, just give up".

Anyway, that's a separate issue. I'll try to look at your patches later, but this week I'm swamped by kernel work.

torvalds:
On Tue, Feb 21, 2023 at 8:25 AM Linus Torvalds
***@***.***> wrote:
>
> I'm in the middle of the kernel merge window, which is a bit
> inconvenient, but this does not look like a libdivecomputer merge
> issue.

Sorry about the long delay - my kernel merge window is finally over,
so I'm looking at the pending issues.

I merged pull request #45 as being trivially fine, but #44 is clearly
fairly contentious, and I'm going to leave that thing alone for now.

However, I *did* revert Jef's change that causes the problem for now,
just to have a libdc state that at least parses the OSTC4 downloads
without giving up. It may not get the gas switches right for those
special cases, but at least the download should work.

I don't have an OSTC4 or a problem case to test with, so this was the
minimal fix I could think of for the current state.

Michael, does that make the current libdc state at least work for you?

                 Linus

mikeller:
Hi Linus.

Unfortunately your revert does not fix the actual problem, as this was pre-existing - importing a CCR dive with gas changes may or may not be aborted with the current implementation, depending on the number of OC bailout gases used during the dive, and the position of the diluents used in the list of diluent gases:

1. `ngasmixes` is set to 5 after parsing of the header on OSTC4;
2. for every different OC bailout gas used during the dive `ngasmixes` is incremented by 1;
3. whenever a switch to closed circuit happens, the index into the list of diluents is is sent with an offset of 5 (on top of the offset of 1 that is used by default, i.e. range [6...10]). If this index is higher than `ngasmixes` the dive import will fail.

So as a consequence of this, 3. will only not fail if enough 2. events have happened beforehand to set `ngasmixes` to a value that is <= the index that is sent in 3.. Since the ordering of the list of diluent gases is arbitrary and in the order that the user is entering them, the success of imports is also arbitrary.

On top of this, even if the dive import does not fail, the gas that is shown in Subsurface for OSTC4 logs as used for closed circuit segments except for the initial one is likely to be wrong, resulting in wrong and potentially misleading information being shown for the decompression status and tissue model.

Cheers
  Michael Keller

torvalds:
On Mon, Mar 6, 2023 at 3:10 PM Michael Keller ***@***.***> wrote:
>
> Unfortunately your revert does not fix the actual problem, as this was
> pre-existing - importing a CCR dive with gas changes may or may not be
> aborted with the current implementation, depending on the number of OC
> bailout gases used during the dive, and the position of the diluents
> used in the list of diluent gases

Ok, it sounded like the problem was new to the merge of Jef's changes.

At the same time, I'm very leery of merging your pr#44, because it
really sounds like doing that

    if (version == 0x24 ..

check is wrong, whether with or without extra checks for divemode or
particular sub-versions of the dive computer.

I'd be *much* happier with a test that literally works on the data
itself, ie something like

#define CC_DILUENT_GAS_OFFSET 5

     diluent = false;
     if (idx > parser->nfixed) {
        idx -= parser->nfixed;
        diluent = true;
     }

where there is no question that you are modifying any situation that
uses the legacy path. So something that *purely* fixes the situation
that currently is an error case.

Hmm?

              Linus

mikeller:
> At the same time, I'm very leery of merging your pr#44, because it
really sounds like doing that
>
>    if (version == 0x24 ..
>
> check is wrong, whether with or without extra checks for divemode or
particular sub-versions of the dive computer.
>
> I'd be *much* happier with a test that literally works on the data
itself, ie something like
>
> #define CC_DILUENT_GAS_OFFSET 5
>
>     diluent = false;
>     if (idx > parser->nfixed) {
>        idx -= parser->nfixed;
>        diluent = true;
>     }
>
>where there is no question that you are modifying any situation that
uses the legacy path. So something that *purely* fixes the situation
that currently is an error case.

I get your point. My proposal is trying to be clinical, and only apply the offset where we know it is needed based on the dive computer (OSTC4) firmware producing the data. Jef has got information and logfiles that show that this offset should never be needed for non-OSTC4 dive computers using this protocol.
But applying the offset solely based on sample data and assuming that whenever we see an index in the [6...10] range this is this case is a possible approach as well - happy to change this particular case in this way. This obviously has the consequence that if such an index is produced by a different case that is a bug in one of the involved firmwares we are failing to report a problem where we should.

Unfortunately, for the second bug that I am addressing in https://github.com/subsurface/libdc/pull/44/files#diff-4004197d7170e5df533d87b83e1b00d179063d881cc7b7de3c784298ad137530R1062-R1068 is of a nature where a fix solely based on the sample data is not possible: For OSTC4 'Extra Gas' changes always change to an OC gas, whereas according to the current code changes to 'Extra Gas' seem to be changing to a diluent (and staying in CC mode) for all other dive computers using this protocol.

I have pushed changes that should be addressing your comments.

PR #47, opened 2023-03-17
Add option to force overwrite firmware for OSTC4.

mikeller:
In order to support development of the open source firmware of the OSTC4.

This is needed in order to be able to install firmware with the same version number multiple times during development and testing.

Breaking API change, needs https://github.com/subsurface/subsurface/pull/3846 to make Subsurface build with it.


torvalds:
Looks good to me, but I just told Dirk to pull this directly, because of the API change.

PR #51, opened 2023-06-14
Fix bugs in OSTC4 support.

mikeller:
Fix two bugs in the support for OSTC4:
- change block size for writes back to 64 bytes to prevent buffer overflow on the dive computer side;
- added 'unsupported' response for the dump command on OSTC4.

### DIFF
--- a/src/hw_ostc3.c
+++ b/src/hw_ostc3.c
@@ -240,7 +240,7 @@ hw_ostc3_write (hw_ostc3_device_t *device, dc_event_progress_t *progress, const 
 	size_t nbytes = 0;
 	while (nbytes < size) {
 		// Set the maximum packet size.
-		size_t length = 1024;
+		size_t length = 64;
 
 		// Limit the packet size to the total size.
 		if (nbytes + length > size)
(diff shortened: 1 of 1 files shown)


glance-:
Changing the max packet size to 64 instead of the previous 1024 makes firmware updates of the OSTC4 ridiculous slow. In what context did you see overflows that prompted this?

jefdriesen:
I'm surprised you notice any difference because it has always been 64 bytes. The change to 1024 bytes was something very recent (commit 1c8cd096b57a876c4fb0afc5113aac05d75d924e).

I assume you are upgrading using rfcomm and not BLE?

mikeller:
@glance-:

> Changing the max packet size to 64 instead of the previous 1024 makes firmware updates of the OSTC4 ridiculous slow. In what context did you see overflows that prompted this?

Yes it is slow indeed - with my configuration Subsurface on ubuntu linux  it has always been, and from testing firmware upload with various block sizes 64 bytes is the maximum block size that will not result in a high percentage of uploads failing due to the dive computer aborting the upload.
What configuration did you use to get faster uploads than this?

glance-:
I might be wrong here, but when upgrading my OSTC4 to 1.6.2 it felt like the upgrade was way slower than what I remembered it to be. It could be that subsurface managed to choose to use btle instead of rfcomm for some reason.

mikeller:
I'll be looking into addressing the upload speed problem in the OSTC4 firmware - once I've got a grip on this we should be able to go faster.

PR #38, opened 2022-06-23
Garmin: Don't require sub directories for fit files

michaelandreen:
Start looking for fit files in the normal Garmin/Activity sub directory,
but if that doesn't exist look directly in the specified directory.

Signed-off-by: Michael Andreen <michael@andreen.dev>

PR #53, opened 2023-07-13
FIT file parser updates

torvalds:
This updates the Garmin FIT file parser to know about developer fields.

Garmin itself doesn't use them (I guess they'd just make them regular fields if they want to do something special), but it looks like Suunto is now generating FIT files as an export from their mobile app. 

This is all somewhat rough to use, since you have to claim you're importing from a Garmin Descent, and set up the Suunto FIT files in the right kind of directory structure (filename less than 24 characters, and in a "Garmin/Activity/" directory), but it does result in what looks like a usable profile.

It's perhaps worth noting that if you download directly from the Suunto EON Core, you get a *lot* more information, and a much better profile, than the FIT export gives you. The Suunto FIT file export really doesn't seem to be very good, but it's better than nothing.


torvalds:
I doubt anybody cares, but I added @dirkhh and @mikeller to reviewers, since you've both at least touched that file.

The biggest question is probably whether we should just move the whole thing into subsurface as a "Import FIT file" thing rather than anything else. Being in libdivecomputer *does* make it convenient for the case of actually plugging in a Garmin dive watch and just mounting it and treating it as other dive computers, but inconvenient for almost everything else.

mikeller (commented):
Looks ok to me - do you think there is anything that we can do to make it possible to fall back to reading from the directory selected by the user as a fallback when the 'Garmin/Activity' directory structure does not exist in the selected location? Seems to be a bit more user friendly, and since we're only picking up .FIT files I don't think this should create problems.

mikeller (approved):

michaelandreen:
> do you think there is anything that we can do to make it possible to fall back to reading from the directory selected by the user as a fallback when the 'Garmin/Activity' directory structure does not exist in the selected location?

That's already the case. Was merged in https://github.com/subsurface/libdc/pull/38

mikeller:
> That's already the case. Was merged in #38

Turns out I've been doing way too much work when testing with .FIT files in the past. 😅

duttabhishek0 (commented):
LGTM!

ghost (commented):
neat program

PR #54, opened 2023-07-25
Cressi Leonardo: Turn ascent warnings into info events.

mikeller:
According to user feedback the ascent warnings on the Cressi Leonardo are very sensitive. They come in 3 different levels. This change turns levels 1 and 2 into info events, leaving only 3 as a warning.
Fixes https://groups.google.com/g/subsurface-divelog/c/qBXF0wmyjKg/m/5qA0_KABCQAJ

Screenshot from 2023-07-25 00-25-22

### DIFF
--- a/src/cressi_leonardo_parser.c
+++ b/src/cressi_leonardo_parser.c
@@ -229,8 +229,13 @@ cressi_leonardo_parser_samples_foreach (dc_parser_t *abstract, dc_sample_callbac
 			if (ascent) {
 				sample.event.type = SAMPLE_EVENT_ASCENT;
 				sample.event.time = 0;
-				sample.event.flags = 0;
+				if (ascent < 3)
+					// Turn the overly sensitive ascent warnings of this dive computer into info events
+					sample.event.flags = SAMPLE_FLAGS_SEVERITY_INFO;
+				else
+					sample.event.flags = 0;
 				sample.event.value = ascent;
+
 				if (callback) callback (DC_SAMPLE_EVENT, sample, userdata);
 			}
 


torvalds:
This doesn't look wrong to me, but I wonder why you also made them string events?

IOW, I think you could have *just* done something like

```
	sample.event.type = SAMPLE_EVENT_ASCENT;
	sample.event.time = 0;
	sample.event.flags = 0;
+	if (ascent < 3)
+		sample.event.flags = SAMPLE_FLAGS_SEVERITY_INFO;
	sample.event.value = ascent;

	if (callback) callback (DC_SAMPLE_EVENT, sample, userdata);
```

instead?

Not a big deal, but considering that Jef hates our string reporting, but *might* be willing to take the severity information flags, it migth be worth limiting the change to just adding the severity bits...

Hmm?

mikeller:
> This doesn't look wrong to me, but I wonder why you also made them string events?

I did this in order to make it possible for users to filter out the ascent info events with 'Hide similar events' in Subsurface, which currently only distinguishes by event name - given the user report that triggered this change this seems to be useful.

An alternative approach might be to use the change you propose above, but then change 'Hide similar events' to also distinguish by value and / or flags. Unfortunately it is hard to tell if this approach will break some other users' workflows.

torvalds:
> 
> An alternative approach might be to use the change you propose above, but then change 'Hide similar events' to also distinguish by value and / or flags. 

That would probably be the right thing to do. In fact, I thought normal events did end up having the strings associated with them already - we definitely do that for things like "gaschange".

Hmm. Looking around, we seem to do that elsewhere too, ie "handle_event()" creates a name for all of these.

So 'Hide similar events' *should* just work. If anything, I think it might be too aggressive (ie it seems to not distinguish by severity).

Anyway, I am not NAK'ing your change.

mikeller:
> Hmm. Looking around, we seem to do that elsewhere too, ie "handle_event()" creates a name for all of these.

Correct - in the Subsurface data model the events just have a name, value, and flags, but the type is implied only in the name.

> So 'Hide similar events' _should_ just work. If anything, I think it might be too aggressive (ie it seems to not distinguish by severity).

Yes, it is too aggressive / not selective enough for this particular use case. With respect to breaking other use cases, do you have an idea which one will be less likely to create problems for other user:
- make 'Hide similar events' distinguish by flags (i.e. warning / info / ...);
- make 'Hide similar events' distinguish by event value (1 / 2 / 3) for 'level' of warning, not sure what it is used for for other dive computers
- or even do both of the above.

> Anyway, I am not NAK'ing your change.

All good, confirmation bias is real and rubber ducking is always appreciated.

torvalds:
So at least in this particular case, I feel like "Hide similar events" should have realized that the severity level information was different, and as such it should not hide *all* ascent warnings, only the "info" ones (if you asked ti to hide an info one, of course).

That's doubly true when to an actual user, if you use only the string to distinguish "similar events", they really look very very different because they use very different visual representations on the profile.

So my gut feel is that it would probably be a good idea to make subsurface take at least the severity level into account when comparing similar events.

But only severity level, not other parts of the event values. For example, for gas changes, the value field contains the gas information you're switching to and the flags contain the cylinder number, so distinguishing on the bases of 'value'or 'flags'  would be wrong.

And in fact, distinguishing on the base of 'type' would also be wrong, since we have two "types" of gas changes for legacy reasons, and they look exactly the same to a user.

But the severity level in the flags makes a rather obvious difference.

(We already hide severity 1 - aka "STATE" - by default and I don't think you can even make them visible, so taking the severity into account makes sense for that reason too).

mikeller:
Sounds good, I'll amend this along these lines, thanks.

torvalds:
This obviously looks fine to me. I assume (or rather - hope) you're doing the subsurface side too?

torvalds:
> I assume (or rather - hope) you're doing the subsurface side too?

Duh. I should have looked closer, you already did.

PR #55, opened 2024-01-21
Garmin: Add Support for Parsing of Garmin Descent Mk3 FIT files.

mikeller:
Add support for FIT files generated by the Garmin Descent Mk3. This just fixes a field number limitation, it has not yet been verified that the Mk3 file format is properly interpreted by Subsurface.

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -30,7 +30,7 @@
 #include "array.h"
 #include "field-cache.h"
 
-#define MAXFIELDS 128
+#define MAXFIELDS 160
 
 struct msg_desc;
 


dirkhh:
It seems silly to ask Linus for a review - but since this is his code... maybe he has other ideas of what to do with the fit file. @mikeller - would you forward that to him as well?

mikeller:
@dirkhh: The only file from a Garmin Descent Mk3 that I have at the moment is for a freedive, which probably isn't very useful for verification of the new sub-format. I'll ping Linus when I've got a FIT file from a SCUBA dive on an Mk3 as well.
From testing with the freedive log, dive time and depth seem to be accurate, and it's just the dive profile that is missing. Since I do not have a freediving test log from an older Garmin Descent model I don't know if this is the expected behaviour for this dive mode.
My idea is to get this change (that seems to be a prerequisite for supporting the Mk3) merged, so that more users can test the import of Mk3 files and report back if there are more issues to be solved.

torvalds:
Looks obviously fine to me.

PR #56, opened 2024-02-07
Import: Update to Latest Upstream Version of libdivecomputer.

mikeller:
Updates included from upstream:

070de23 Post release version bump to 0.9.0
a34e909 Change the units for the sample time to milliseconds
bca8f9e Enable the millisecond resolution sample time
b1ff2c6 Add the sensor index to the ppO2 sample
4b383a7 Add a TTS field to the deco sample
becb8bd Add a usage field to the tank and gas mix
4e24b3a Pass the sample struct by reference
0a4f377 Remove the backend specific calibration functions
679db0b Remove the clock parameters from the constructor
63f5a4d Remove the dc_parser_set_data function
9b7aa81 Change the salinity format in the xml output
ee147af Move the model number mapping to a function
13705f2 Use a static buffer for the RDBI function
d4402aa Add support for the WDBI function
c16530b Detect negative response packets
25bd1f9 Add support for time synchronization
ff03285 Ignore the diluents for open-circuit dives
3a68af4 Fix the OSTC4 firmware upgrade
0afd62d Return an error for the OSTC4 memory dump
ceaaba3 Add support for the new Ratio iX3M 2 models
f818a5a Add a function for detecting CCR dives
f77e9c0 Restrict the oxygen/diluent usage to CCR dives
a4cd21b Use the GTR mode to detect sidemount tanks
9bc742d Use the HP CCR data for the oxygen/diluent usage
8bfbb94 Include the command byte in the hexdump
ecc9e0b Limit the lookup function to the manual gas mixes
323804d Keep track of the actively used gas mixes
993283d Store the original one based gas mix index
5d36cc0 Remove disabled gas mixes
301fdbe Remove support for the Tusa TC1
b2310e6 Pass the descriptor to the filter function
a985b11 Replace the filter parameters with an alternative
4a9be44 Use separate data structures for USB and USB HID
ed87113 Export the filter function in the public api
fe9b47f Document the device descriptor functions
330cb88 Use plain ascii text in the code comments
baa1c49 Use a symbolic constant for the header size
072f0d4 Fix a potential buffer overflow
a7e7439 Setup the 2 byte command header internally
e83732e Fix the Mares Bluelink Pro communication
3c50e91 Allow a zero timestamp for the first sample
e1762fc Skip all non-sample records
2d9008a Remove disabled gas mixes
08d8c3e Add support for parsing the compass bearings
d47e1ce Add support for the Scubapro G3 and Luna 2.0
cfe345a Exclude O2 sensors without calibration data
de6696b Add support for the Shearwater Tern
1d0aeec Fix the React Pro White memory layout

Plus fixes for Garmin Descent devices and fixes to make gasmixes wor for CCR dives in Subsurface.


dirkhh:
Sweet

What's the plan? We merge and use in latest builds and see what fails?

mikeller:
The changes do need corresponding changes in Subsurface due to the API changes in libdivecomputer. I do have these ready to go in https://github.com/mikeller/subsurface/tree/update_libdivecomputer_202402, so next will be to add an update of the submodule changeset to this branch and open it as a pull request.
This has also had limited testing with the few dive computers that I have available.

dirkhh:
And that is exactly how things have been done in the past. Merge libdc, merge that and the corresponding changes into Subsurface, and then wait for people to run into problem (after testing with those DCs you have laying around)...

Ironically, I just dug out my Peregrine to make sure that still worked :)

mikeller:
Happy to do more of this testing if a manufacturer wants to send me test hardware. 😝

jefdriesen:
> Happy to do more of this testing if a manufacturer wants to send me test hardware. 😝

In most cases, you don't really need physical access to a dive computer for testing. Unless the problem is in the transport layer, you can get very far with the simulator and replaying logs. For the parser, the raw dive data is all you need. That's how I test most of the libdivecomputer code.

jefdriesen:
What's the reason for splitting the usage enums, and removing the oxygen/diluent from the tank enum? This actually removes valuable information, especially for dive computers like the Shearwaters that provide no link between the tanks and its gas mix.

You have a TODO comment in the Shearwater code to "Link this tank to the correct gasmix", but unfortunately that is not possible because the information to do that is simply not there. You can guess at best, and in many cases the result will be correct (especially for simple scenarios which are luckily also the most common ones), but it'll never be 100% reliable. There are already several bug reports where subsurface gets this wrong.

In some cases, the usage info can help to make a better guess, and that's exactly why I added it. But this information gets removed in your changes.

Let me illustrate with a real life example from a Shearwater Nerd in HPCCR mode. The dive computer reports 3 gas mixes:

```
<gasmix>
   <he>0.0</he>
   <o2>50.0</o2>
   <n2>50.0</n2>
</gasmix>
<gasmix>
   <he>0.0</he>
   <o2>21.0</o2>
   <n2>79.0</n2>
</gasmix>
<gasmix>
   <he>27.0</he>
   <o2>19.0</o2>
   <n2>54.0</n2>
   <usage>diluent</usage>
</gasmix>
```

And 4 tanks:

```
<tank>
   <beginpressure>199.81</beginpressure>
   <endpressure>199.67</endpressure>
</tank>
<tank>
   <beginpressure>182.57</beginpressure>
   <endpressure>178.85</endpressure>
</tank>
<tank>
   <usage>diluent</usage>
   <beginpressure>181.06</beginpressure>
   <endpressure>108.94</endpressure>
</tank>
<tank>
   <usage>oxygen</usage>
   <beginpressure>150.03</beginpressure>
   <endpressure>104.94</endpressure>
</tank>
```

First problem is that the number of gas mixes and tank does not match. That's because the oxygen tank does not have a corresponding gas mix entry. In HPCCR mode, the dive computer indicates the last two tanks are the diluent and oxygen tank. With this information you could link the diluent tank to the TX 19/27 diluent mix and add the missing O2 mix, and the result would be correct for this dive. Without the tank usage info you can't really do this.

The next problem is that we still have two tanks and two gas mixes left. These are bailout tanks, which were not used throughout the dive. So there are no gas switches or pressure changes that might give a clue which tank and gas mix were used a the same time and could be linked to each other. There is absolutely no way to get this right in all cases, because the only source of truth is the user here! In this case, a simple one-to-one mapping will produce the wrong result because the first tank contained AIR, and the second one EAN50 (confirmed by the user). Note that the user has no control over this, because the Shearwater computers always orders the gas mixes by their oxygen content.

For the non HPCCR cases, libdivecomputer relies on the tank name to guess the oxygen/diluent usage ('O' for Oxygen, and 'D' for Diluent). This is not 100% reliable, but at least it does give the user a chance to configure his computer such that the info is correct.

mikeller:
@jefdriesen:
> What's the reason for splitting the usage enums, and removing the oxygen/diluent from the tank enum?

To clean up the code and make it clear what information is provided in which place by the dive computer.

> This actually removes valuable information, especially for dive computers like the Shearwaters that provide no link between the tanks and its gas mix.

Apart from the edge case of a CCR model running custom firmware on a Shearwater device I could not find any place where my change removes factual information coming from the dive computer. And this edge case is marked in a TODO, and it should be possible to cover this on Subsurface side once proper cylinder management has been implemented.

The key to fixing the shortcomings in cylinder management in the current version of Subsurface will be to implement functionality that allows the user to map gasmixes to tank sensors (or even to tanks they created themselves if they are not using sensors).

PR #58, opened 2024-04-21
Garmin: Add Support for Descent Mk3(i).

mikeller:
Add support for the Garmin Descent Mk3(i) models.

This pretty much just updates the entry in the device list to cover Mk2 and Mk3 models.
It does nothing to add support for MTP when reading from an Mk3 model - I have not been able to get MTP support on linux going reliably, and I do not have an Mk2 / Mk3 to test with anyway.
I suspect that MTP support is incomplete anyway, as the product IDs for models like the Mk2S and Mk2G are not detected.

This change also adds an extra info field for the Model (based on a heuristic and incomplete list).
@torvalds do you have any contact information for Garmin - asking them for a list of model IDs (and USB product IDs) is probably the best approach to get this improved.
I have also added more detailed debug information that could help with guessing the USB product IDs for Descent models that support MTP.

It also fixes the data type for the sensor_type field.

### DIFF
--- a/src/descriptor.c
+++ b/src/descriptor.c
@@ -471,10 +471,11 @@ static const dc_descriptor_t g_descriptors[] = {
 
 	// Not merged upstream yet
 	/* Garmin -- model numbers as defined in FIT format; USB product id is (0x4000 | model) */
-	/* for the Mk1 we are using the model of the global model - the APAC model is 2991 */
-	/* for the Mk2 we are using the model of the global model - the APAC model is 3702 */
+	/* for the Mk1 we are using the model of the global model */
+	/* for the Mk2/Mk3 we are using the model of the Mk2 global model */
+	/* see garmin_parser.c for a more comprehensive list of models */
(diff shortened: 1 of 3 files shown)


dirkhh:
I reached out to Garmin to get the missing IDs.
They have been supportive in the past, but my previous contact left the company so I'm going through the motions again to re-establish the connection

mikeller:
More changes:
Ignore timestamps on messages that we do not read any other fields from
- with the new (firmware >= 16.x) .fit file format these messages seem
to be sent out of order, causing other messages to be ignored if they
come after them in the file but are before them in the timeline.
Track the OC / CCR state at the beginning of the dive based on the
dive type - this avoids spurious 'switched to ...' messages caused by
new 'gas switch' messages at the beginning of dives.

See also https://forums.garmin.com/developer/fit-sdk/b/news-announcements/posts/important-fit-activity-file-message-change

PR #61, opened 2024-05-04
Shearwater: Add Stack Timer Information.

mikeller:
Add the configured stack run time and the remaining stack run time at the start and end of the dive as extra information.

image

### DIFF
--- a/src/shearwater_predator_parser.c
+++ b/src/shearwater_predator_parser.c
@@ -497,6 +497,8 @@ shearwater_predator_parser_cache (shearwater_predator_parser_t *parser)
 
 	unsigned int offset = headersize;
 	unsigned int length = size - footersize;
+	unsigned int stack_time_total_s = 0;
+	unsigned int stack_time_remaining_s;
 	while (offset + parser->samplesize <= length) {
 		// Ignore empty samples.
 		if (array_isequal (data + offset, parser->samplesize, 0x00)) {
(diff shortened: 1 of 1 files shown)


glance-:
Nice. I Just started 3 days ago looking into the data for these values! :+1:

glance-:
Are you certain that these values are unsigned? Have you looked at these values when you go over your configured stack time? I'm reasonably certain that I've seen stack-time gone negative on the computer when you've exceeded your configured stack time.

I think I might have some such values in my computer and I'll play around with this later this week.

glance-:
Ok, I actually got around to look at this tonight. Time remaining is just 0 when you've gone over your configured stack time. I haven't manage to view this data either on the computer or in the shearwater app. It's probably to new feature for them to have implemented it.

The only thing I can think of is that the key strings are quite long. I'm not sure if describing the format "[hh:mm:ss]" is necessary, and the "stack run time" is longer than the name shearwater uses, which is just "stack time".

I'd vote for "Configured stack time", "Stack time at start" and "Stack time at end".

mikeller:
Thanks for checking the 'going negative' part.
The stack timer on the Shearwater starts giving a non-suppressable warning whenever the remaining stack time goes under 30 minutes, which to me is not desirable during the last minutes of decompression, so I always configure my stack timer to be desired maximum stack run time plus 30 minutes, and consequently have never come close to 0 on the timer.

Re Extra Info Key names, there is a lot of space in the Extra Info tab, and the density of information displayed there is relatively low compared to other tabs, so I figure we may as well use it for key names. But I'll have a think about changing the names.

image

I also think it pays to be concise and add units to values that are returned, and getting a 'xx:yy' value returned (which is how this started out before I discovered that Shearwater is logging seconds for the stack run time), and not knowing if this is hh:mm or mm:ss is one annoying thing that you run into if there is no conciseness about units.

mikeller:
I ended up going for 'Configured stack time', 'Remaining stack time at start', and 'Remaining stack time at end', as to me 'xxx time' seems to imply that this is a dimension that is counting up, whereas 'remaining xxx time' clarifies that it is counting down.

src/shearwater_predator_parser.c
```diff
@@ -704,6 +714,14 @@ shearwater_predator_parser_cache (shearwater_predator_parser_t *parser)
 				}
 			}
 		} else if (type >= LOG_RECORD_CLOSING_0 && type <= LOG_RECORD_CLOSING_7) {
+			if (type == LOG_RECORD_CLOSING_6) {
+				if (logversion >= 11) {
+					if (stack_time_total_s > 0) {
+						stack_time_remaining_s = array_uint16_be(data + offset + 3);
+						dc_field_add_string_fmt(&parser->cache, "Remaining stack time at end [hh:mm]", "%d:%02d:%02d", stack_time_remaining_s / 3600, (stack_time_remaining_s / 60) % 60, stack_time_remaining_s % 60);
```
glance-: Shouldn't this be hh:mm:ss as the others?

mikeller: Correct, thanks, fixed.

glance- (approved):

PR #66, opened 2024-07-27
Update libdivecomputer with changes in upstream as of 20240725

mikeller:
Tested with Subsurface, all seems to be working fine.


mikeller:
I just discovered that this is somehow breaking Garmin support.

mikeller:
My bad, it's just non-helpful UX and me leaving bluetooth checked when testing.

PR #68, opened 2024-09-09
Remove Devices not Supported by Subsurface.

mikeller:
Remove the latest generation of devices made by Pelagian Pressure Systems, as their intentionally broken pairing mechanism is not supported by Subsurface.

Note: This list of devices is only used by the mobile version.

image


dirkhh (approved):
That makes sense to me

PR #69, opened 2024-12-26
Garmin: Report correct local time offset

torvalds:
We stupidly thought that the local time offset was in the "DEVICE_SETTINGS" message as the time_offset field.  I'm pretty sure I've seen something like that before.

But the FIT files from github issue #4401 clearly have that time offset field being zero, and there are two other ways to figure out what local time actually is, namely in the "local_time" field of the ACTIVITY message or the TIMESTAMP_CORRELATION message.

Either of those seem to work for what we want, so let's parse both (we already did the ACTIVITY case), and let's ignore the "time_offset" field from DEVICE_SETTINGS at least if it is zero.

There is probably some real explanation for what the proper way to deal with all this is, and what the whole time_offset and utc_offset fields from DEVICE_SETTINGS means, and maybe we can improve on this in the future if somebody figures it all out.

In the meantime, this seems to be an improvement.

Reported-by: @WetsuitSeasoning
Link: https://github.com/subsurface/subsurface/issues/4401

### DIFF
--- a/src/garmin_parser.c
+++ b/src/garmin_parser.c
@@ -617,8 +617,23 @@ DECLARE_FIELD(RECORD, rmv, UINT16) { }			// 100 * l/min
 DECLARE_FIELD(RECORD, ascent_rate, SINT32) { }		// mm/s (negative is down)
 
 // DEVICE_SETTINGS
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
This looks reasonable and doesn't appear to break the test downloads that I did with it.

t00t4:
é isso ai papai

PR #71, opened 2024-12-30
Update libdivecomputer from Upstream.

mikeller:
Add the latest updates in libdivecomputer as of 2024/12/31:

- **Add support for a new variant of the Aqualung i100**
- **Add a new Perdix 2 hardware ID**
- **Add support for the i330R Console**
- **Return the correct error code**
- **Don't skip the last entry in the logbook buffer**
- **Add ioctl's for reading and writing BLE characteristics**
- **Add helper functions to convert UUID to/from strings**
- **Add support for the Cressi BLE protocol variant**
- **Remove the unused model parameter**
- **Add the device id and logbook entry**
- **Add an extra surface sample**
- **Add timesync support**
- **Exit PC Link mode when operation is done**
- **Add support for the Cressi Nepto**
- **Add support for the Cressi Leonardo 2.0**
- **Add CCR support for the Excursion**
- **Add support for the Uwatec Aladin One**

### DIFF
--- a/contrib/android/Android.mk
+++ b/contrib/android/Android.mk
@@ -9,6 +9,7 @@ LOCAL_SRC_FILES := \
 	src/array.c \
 	src/atomics_cobalt.c \
 	src/atomics_cobalt_parser.c \
+	src/ble.c \
 	src/bluetooth.c \
 	src/buffer.c \
 	src/checksum.c \
--- a/contrib/msvc/libdivecomputer.vcxproj
+++ b/contrib/msvc/libdivecomputer.vcxproj
@@ -177,6 +177,7 @@
     <ClCompile Include="..\..\src\array.c" />
     <ClCompile Include="..\..\src\atomics_cobalt.c" />
     <ClCompile Include="..\..\src\atomics_cobalt_parser.c" />
+    <ClCompile Include="..\..\src\ble.c" />
     <ClCompile Include="..\..\src\bluetooth.c" />
     <ClCompile Include="..\..\src\buffer.c" />
     <ClCompile Include="..\..\src\checksum.c" />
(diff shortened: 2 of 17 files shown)


dirkhh:
I assume that a number of these (at least the BLE one) are no-ops for us. Not saying they shouldn't be merged, just maybe that should be noted in the commit message?
We need to remember to run the scripts to update the list of supported dive computers when this is merged in Subsurface. And IIRC, we may need to manually remove the ones that aren't supported using Qt BLE without special handling for their broken-by-design protocols.

torvalds (approved):
Looks all reasonable to me

mikeller:
> I assume that a number of these (at least the BLE one) are no-ops for us. Not saying they shouldn't be merged, just maybe that should be noted in the commit message?

All of the commits in this pull request (except for the merge commit) are the result of a `git merge` of libdivecomputer's `master` branch - so I am not sure if it makes sense to go and edit these commit messages and make them different from the ones in the 'upstream' repository?

> We need to remember to run the scripts to update the list of supported dive computers when this is merged in Subsurface.

Yes, I'll do that. We  can probably improve the process by adding a GitHub workflow that checks for every pull request if it will require the scripts to be run, and fails otherwise.

And while we are at it, I think there is generated documentation that is kept in the repository. Maybe this should have a similar guard?

> And IIRC, we may need to manually remove the ones that aren't supported using Qt BLE without special handling for their broken-by-design protocols.

This already came out in the wash when resolving merge conflicts:

https://github.com/subsurface/libdc/pull/71/files#diff-470ccbf1ede979acb859b72fe3f44f1ff0d70a5599224b3c0097dc9e2a107487L279-L283

mikeller:
Tested with:
- Shearwater Petrel 2
- Heinrichs Weikamp OSTC4
- Garmin Descent Mk1

PR #72, opened 2025-01-01
Garmin: Remove File Name from File Data.

mikeller:
Remove the file name from the beginning of the file data buffer. This is
in preparation to a move to shift the .FIT file importer to 'Import log
files'.
The file name in the buffer does not seem to be used for anything, and
it seems to be wrong to have this added only for the parser to then have
to remove this non-file-content again first thing.
Fingerprinting has been confirmed as working after this change.

@torvalds, @dirkhh: As the original creators of the Garmin importer, do you know if there are any hidden reasons to have the file name included in the file content buffer?

### DIFF
--- a/src/garmin.c
+++ b/src/garmin.c
@@ -558,7 +558,6 @@ garmin_device_foreach (dc_device_t *abstract, dc_dive_callback_t callback, void 
 
 		// Reset the membuffer, read the data
 		dc_buffer_clear(file);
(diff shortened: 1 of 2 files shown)


mikeller:
Apologies @torvalds and @dirkhh for merging this before you could review it, but due to a broken git hook I accidentally pushed 5889c1a3f82d21d56cf33b62e72e73af6c5a2835 into https://github.com/subsurface/subsurface, and this needed to go in to make the .FIT log file import work.

PR #76, opened 2025-04-07
Add support for a new Sirius profile version

rmultan:
After upgrading a Mares Sirius to firmware version 1.6.16, the profile version changed to v2.0. Surprisingly, despite the change in the major version number, there appear to be no functional changes in the data format.


mikeller:
This seems to be an identical change to the one in 82712ce8f83b2448d01ad26a949a82c27452ba5d. This will create all kinds of problems the next time updates from the upstream libdivecomputer are merged into our 'fork', unless it is done in a cherry-pick.
I'll pull this in in a cherry-pick when I get around to.

PR #86, opened 2025-06-08
Fix Scubapro Luna 2.0 sample interval to display correct dive duration.

anttisep:
With this fix dives imported from Scubapro Luna 2.0 to Subsurface-mobile will display correct dive durations. With default interval of 4, dive durations will be twice as long as they should be. Reducing interval to 2 fixes this.

### DIFF
--- a/src/uwatec_smart_parser.c
+++ b/src/uwatec_smart_parser.c
@@ -916,6 +916,11 @@ uwatec_smart_parse (uwatec_smart_parser_t *parser, dc_sample_callback_t callback
 	double density = (parser->watertype == DC_WATER_SALT ? SALT : FRESH);
 
 	unsigned int interval = 4;
+	
+	if (parser->model == LUNA2) {
+		interval = 2;
+	}
+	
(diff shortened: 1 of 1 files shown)


mikeller:
From the discussion in https://groups.google.com/g/subsurface-divelog/c/TpcKuejb_Q8/m/HS017JTLAAAJ, the information provided by Scubapro seems to indicate that the change to the sample interval was introduced by them in error, and they are planning to revert this in the next firmware update.
So changing this in the way you are proposing will break this now for all users who have not yet updated their firmware, and will break it for all users who apply the next firmware update once it is available. All in all I am not sure if this fix isn't more disruptive than just 'sitting it out' - or alternatively a 'future proof' fix could be done by checking the firmware version, and only using the changed interval if it matches the buggy version.

anttisep:
Thank you mikeller for pointing out this thread! I did spent quite some time for googling the problem after switching to my new dive computer last week, but my google magic just wasn't at the proper level to find it :)

You are absolutely correct and this fix would even make things worse and confusing. I'll try to see if it is in anyway within my capabilites to do the check, but for now I'll close the PR.

PR #94, opened 2025-12-13
Fix garmin tts parsing

dotanalon:


mikeller (changes requested):
This is incomplete. See my comments on https://github.com/subsurface/subsurface/pull/4654.

Also, have a look at fixing the DCO: https://github.com/subsurface/libdc/pull/94/checks?check_run_id=57976947694

PR #95, opened 2025-12-14
Fix garmin tts parsing (fixed DCO)

dotanalon:


mikeller:
@dotanalon: This is the same fix as what you had proposed in #94, which you then closed before opening this. If you want to discuss any more you can re-open #94 and implement the changes requested.
Please refrain from opening more unneeded pull requests.

dotanalon:
Unfortunately I cant reopen it, what do you suggest to do?

PR #99, opened 2026-01-13
Separate the Reporting of 'bearing' vs. 'heading.

mikeller:
There seems to be two different ways that dive computers report
directional information:

1. report headings that are set by the user by pointing the device in
   the desired direction, then pressing a button;

2. report the direction that the device is pointing in at set intervals
   (like every 10 seconds) during the dive.

For a user-friendly visualisation, these two approaches need to be shown
in different ways in Subsurface:

1. show an event in the profile every time the user executes the 'set
   heading' operation (and potentially at the start of the dive for the
   initial heading):

image

2. show the current bearing in the information box in the profile when
   the mouse pointer traverses the profile:

image

Subsurface will be using different ways for libdivecomputer backends to
import this data:

1. using `DC_SAMPLE_EVENT` samples with `SAMPLE_EVENT_HEADING` type;

2. using `DC_SAMPLE_BEARING` samples.

This changeset updates the following dive computers to use 1. to match
their type of reporting:

- Heinrichs Weikamp OSTC3 / 4 / 5
- ScubaPro / Uwatec Smart

The following dive computers are already using 1. correctly:

- Suunto EON Steel
- Suunto D9

And the following dive computer seems to be using 2. correctly:

- Divesoft Freedom / Liberty (@jakubhecko are you able to confirm this?)

The following dive computers are currently using 2., but we don't have
any information if this is correct:

- DiveSystem iDive
- Shearwater Predator
- Halcyon Symbios

### DIFF
--- a/src/hw_ostc_parser.c
+++ b/src/hw_ostc_parser.c
@@ -1184,7 +1184,6 @@ hw_ostc_parser_internal_foreach (hw_ostc_parser_t *parser, dc_sample_callback_t 
 
 				if (callback) {
 					unsigned int value = array_uint16_le(data + offset);
-					dc_sample_type_t eventType = DC_SAMPLE_EVENT;
 					dc_sample_value_t sample = {
 						.event.type = SAMPLE_EVENT_STRING,
 						.event.flags = SAMPLE_FLAGS_SEVERITY_INFO,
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
looks reasonable

PR #100, opened 2026-01-20
Garmin: Add Timezone Offset Information to Extra Info.

mikeller:
Add the offset of the reported dive time to UTC as an Extra Info field.

Note that the same offset can also be extracted from the `timezone` field
that is returned by `dc_parser_get_datetime()`.

image


bstoeger:
Hmmm... If this information is already in `struct dc_datetime_t`, I see no reason to add it as extra_info. Let's just use what is there?

mikeller:
> Hmmm... If this information is already in `struct dc_datetime_t`, I see no reason to add it as extra_info. Let's just use what is there?

That's what I thought too, which is why I opened subsurface/subsurface#4689 to supersede this. Let's discuss there.

PR #106, opened 2026-02-25
Msvc usb storage

dirkhh:
trying to see if copilot will offer a review if I open the PR from a different repo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
           sudo apt-get install gcc-mingw-w64 binutils-mingw-w64 mingw-w64-tools
       - name: Install libusb
         env:
-          LIBUSB_VERSION: 1.0.26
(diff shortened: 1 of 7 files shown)


mikeller:
Is this still needed?

dirkhh:
apparently not. I'm not quite sure why I needed the changes to garmin_parser.c -- or more likely which settings I was missing when I did this locally that prevented the existing code from working and caused me to need the changes in this PR, but since the GitHub Action appears to successfully build libdc without these changes, I guess... 🤷🏼‍♂️

The bigger issue for Garmin is that we don't have a working libmtp for Windows/msvc - but that has nothing to do with this PR.

mikeller:
> The bigger issue for Garmin is that we don't have a working libmtp for Windows/msvc - but that has nothing to do with this PR.

Ouch, yes, that would be a loss, even more so as Garmin is only supported on desktop. How ugly will it get if we have to use the native Windows MTP API?

PR #111, opened 2026-05-11
Shearwater: Fix the Firmware Version Decoding.

### LINKED ISSUE #4806: [Bug]: Shearwater firmware version displayed incorrectly (hex BCD) for versions >= 100
**Description**

Shearwater dive computers (e.g. Perdix 2) with firmware version 100 or higher display the firmware version incorrectly in Subsurface's **Extra Info** tab. Instead of showing the decimal firmware version (e.g. `100`, `102`), it shows the raw hex BCD-encoded byte value (e.g. `a0`, `a2`).

**Background:** This was first discussed in #4537, which tracked a related connection failure for Perdix 2 on firmware v100 (fixed in `libdivecomputer`). The firmware version display issue in the parser was noted as a separate follow-up there.

The firmware version is stored as a BCD (Binary-Coded Decimal) encoded byte in the dive log data (byte 19 of the Shearwater parser). BCD normally represents values 0–99 cleanly. When Shearwater reached firmware v100, the high nibble exceeded the standard BCD range — `0xA0` = v100, `0xA1` = v101, `0xA2` = v102, etc. The current code formats this byte as raw hex (`%2x`), causing it to display as `a0` or `a2` instead of `100` or `102`.

**Subsurface Version**

- macOS: 6.0.5605
- iOS: 6.0.5581

**Platform(s)**

- iOS, macOS

**Platform Version**

- iOS 18.4.2
- macOS 15.4.1

**Divecomputer Model**

Shearwater Perdix 2

**Steps to Reproduce**

1. Connect a Shearwater Perdix 2 running firmware v100 or later
2. Download dives in Subsurface
3. Open the **Extra Info** tab for any dive
4. Observe the **FW Version** field shows `a0` (for v100) or `a2` (for v102) instead of the correct decimal value

**Proposed Fix**

In `shearwater_predator_parser.c`, decode the BCD byte properly instead of printing as raw hex:

```c
dc_field_add_string_fmt(&parser->cache, "FW Version", "%u", bcd2dec(data[19]));
```

Suggested by @jefdriesen in #4537. Note that @mikeller was waiting for Shearwater to clarify the exact encoding, but the general BCD decoding approach is likely correct.

**References**

- Original related issue: #4537
- libdivecomputer length fix: https://github.com/libdivecomputer/libdivecomputer/commit/05f510416eda088e57382b4955bf53ee919dbf58

**Mentions**

@jefdriesen @mikeller

mikeller:
Decode the Shearwater firmware version as BCD.

Fixes https://github.com/subsurface/subsurface/issues/4806.

### DIFF
--- a/src/shearwater_predator_parser.c
+++ b/src/shearwater_predator_parser.c
@@ -897,7 +897,7 @@ shearwater_predator_parser_cache (shearwater_predator_parser_t *parser)
 
 	dc_field_add_string_fmt(&parser->cache, "Serial", "%08x", parser->serial);
 	// bytes 1-31 are identical in all formats
-	dc_field_add_string_fmt(&parser->cache, STRING_KEY_FIRMWARE_VERSION, "%2x", data[19]);
+	dc_field_add_string_fmt(&parser->cache, STRING_KEY_FIRMWARE_VERSION, "%u", bcd2dec(data[19]));
 	add_deco_model(parser, data);
 	add_battery_type(parser, data);
 	dc_field_add_string_fmt(&parser->cache, "Battery at end", "%.1f V", data[9] / 10.0);


mikeller:
This seems to be working well - but then I do not own a Shearwater dive computer that can run a firmware version newer than 97.

dirkhh:
It cracks me up that this is considered bcd

A is not a decimal 🤣

mikeller:
> It cracks me up that this is considered bcd
> 
> A is not a decimal 🤣

Yeah, makes one wonder if we'll ever get to `0xaa`, or if this will be `0xb0`, or what...

PR #117, opened 2026-07-11
Add support for the Shearwater Perdix 3 BLE protocol

splichy:
The Shearwater Perdix 3 keeps the classic Shearwater application protocol
(SLIP, RDBI/WDBI, block download, same register IDs and dive parser) but
changes two things at the BLE transport layer. This adds a framing switch
to the shared Shearwater/Petrel code so both the classic devices and the
Perdix 3 are handled by one implementation.

### What changes on the Perdix 3

- **New GATT service.** It does not expose the classic Shearwater serial
  service; it advertises a new vendor service. Detection is by the BLE
  name `Perdix 3` (matched before `Perdix`, since matching is by prefix).
- **No per-write frame prefix.** Classic BLE framing prefixes every write
  with a 2-byte `[nframes, index]` header. The Perdix 3 sends a raw SLIP
  stream with no per-packet header; notifications are reassembled until
  the SLIP END byte `0xC0`.
- **24-bit transfer length.** The transfer header carries a 24-bit
  big-endian payload length (`FF 01 <len24>` request, `01 FF <len24>`
  response) instead of the classic 8-bit length.
- **Wider download blocks.** The download init uses record type `0x24`
  with a 16-bit size field and the init response is `75 20 <blocksize:16be>`.
  Observed block sizes are 514 bytes (manifest) and 2306 bytes (compressed
  dive), so the packet buffers are raised from 254 to 4096 bytes.

### Implementation

A `framing` field on `shearwater_common_device_t`
(`SHEARWATER_FRAMING_CLASSIC` / `SHEARWATER_FRAMING_PERDIX3`) selects the
behavior. It is derived from the descriptor model, which is now passed into
`shearwater_petrel_device_open`. A new descriptor row maps the `Perdix 3`
BLE name to a new model and maps hardware type `0x39C2` to it. The classic
path is unchanged.

### Testing

Verified end-to-end against a Perdix 3 (firmware V38) over BLE on Linux:
identity reads (serial/firmware/hardware), manifest read, compressed dive
download, and full dive import all succeed. The same build cross-compiled
for Android arm64 was validated end-to-end on a physical phone.

### Notes for reviewers

Portions of the new protocol logic were AI-assisted; the non-trivial new
blocks in `shearwater_common.c` are marked with an `AI-generated (Claude)`
comment. The protocol was reverse-engineered from an HCI snoop of the
official Shearwater app and byte-matched against the capture.

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/src/descriptor.c
+++ b/src/descriptor.c
@@ -392,6 +392,7 @@ static const dc_descriptor_t g_descriptors[] = {
 	{"Shearwater", "Tern",      DC_FAMILY_SHEARWATER_PETREL, 12, DC_TRANSPORT_BLE, dc_filter_shearwater},
 	{"Shearwater", "Tern TX",   DC_FAMILY_SHEARWATER_PETREL, 12, DC_TRANSPORT_BLE, dc_filter_shearwater},
 	{"Shearwater", "Peregrine TX", DC_FAMILY_SHEARWATER_PETREL, 13, DC_TRANSPORT_BLE, dc_filter_shearwater},
+	{"Shearwater", "Perdix 3",  DC_FAMILY_SHEARWATER_PETREL, 14, DC_TRANSPORT_BLE, dc_filter_shearwater},
 	/* Dive Rite NiTek Q */
 	{"Dive Rite", "NiTek Q",   DC_FAMILY_DIVERITE_NITEKQ, 0, DC_TRANSPORT_SERIAL, NULL},
 	/* Citizen Hyper Aqualand */
@@ -764,6 +765,7 @@ dc_filter_shearwater (const dc_descriptor_t *descriptor, dc_transport_t transpor
 		"NERD 2",
 		"Perdix",
 		"Perdix 2",
+		"Perdix 3",
 		"Teric",
 		"Peregrine",
 		"Peregrine TX",
--- a/src/device.c
+++ b/src/device.c
@@ -210,7 +210,7 @@ dc_device_open (dc_device_t **out, dc_context_t *context, dc_descriptor_t *descr
 		rc = shearwater_predator_device_open (&device, context, iostream);
 		break;
 	case DC_FAMILY_SHEARWATER_PETREL:
-		rc = shearwater_petrel_device_open (&device, context, iostream);
+		rc = shearwater_petrel_device_open (&device, context, iostream, dc_descriptor_get_model (descriptor));
 		break;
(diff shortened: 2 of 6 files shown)


splichy:
@mikeller

mikeller:
Making this change directly in the Subsurface 'fork' of libdivecomputer is not the best way to get this done - this wil create friction when merging changes from uptream libdivecomputer in the future.
So it will be better if we wait until libdivecomputer has added support for the Perdix 3, and then cherry-pick that change. Feel free to ping me when this has happened.

Also note that this is an open source project, and maintainers will volunteer their free time to work on it whenever they have it available, and have the motivation to do so. Pinging them within hours of submitting something will not help with motivation...

splichy:
Ooh, didn't know it's a fork :) 
I see there is an active PR for Perdix 3 support in upstream, and the implementation seems roughly the same.

mikeller:
Yeah, it's an unofficial fork of libdivecomputer - we regularly pull in updates from upstream, but we also have some improvements (like support for sending string messages to the host applications, or file / MTP based devices) and bugfixes that were never accepted into upstream.

splichy:
Closing this. Upstream libdivecomputer has now merged its own Perdix 3 support (`8e564eb5`, "Merge branch 'perdix3'"), so the hand-rolled commit in this PR is superseded.

Per @mikeller's note that `Subsurface/libdc` should pick up Perdix 3 by cherry-picking from upstream rather than taking a bespoke implementation, the right path is to bring in the four upstream commits instead:

- `24a68b79` Use symbolic constants for the commands
- `661cb6e5` Handle variable sized packets
- `b9240920` Add Shearwater Perdix 3 support
- `06120b0a` Update the minimum and maximum BLE packet size

I test-cherry-picked all four onto `Subsurface-DS9` (`8717731`) locally - they apply cleanly with no conflicts, and the resulting `descriptor.c` / `shearwater_common.*` / `shearwater_petrel.*` files are byte-identical to upstream master for the touched regions. Happy to open a clean cherry-pick PR if that's useful, or leave it to the normal upstream-sync cadence - whichever you prefer.

The Subsurface-side transport change (service UUID + scan name) is tracked separately in subsurface/subsurface#4864; it stays necessary regardless, since upstream selects the v2 protocol purely by model number and carries no knowledge of the new GATT service UUID.

PR #118, opened 2026-07-14
Add Support for the Shearwater Perdix 3.

mikeller:
Cherry picked from upstream.


splichy:
Tested this cherry-pick end-to-end against a physical Shearwater Perdix 3.

Built both the Linux desktop and Android (arm64) apps with the libdivecomputer submodule checked out at this PR's head (`ab18ece6`), combined with the Subsurface transport-layer change tracked in subsurface/subsurface#4864 (the new GATT service UUID `1aa44039-...` in `qt-ble.cpp` plus the scan-name entry in `btdiscovery.cpp`, which are still required since the upstream implementation selects the V2 protocol by model number and carries no knowledge of the service UUID).

On both platforms the Perdix 3 connects over BLE, downloads its dive log, and imports correctly, including dive-computer time sync. No conflicts, no code changes needed on top of this PR.

